PR #20, opened 2024-07-05
Add `dev` and `dev_native` features

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy-template/issues/18.

`bevy/dynamic_linking` works for both web and native builds, so it goes in `dev`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "bevy_github_ci_template"
+name = "bevy_template"
 version = "0.1.0"
 edition = "2021"
 license = "MIT OR Apache-2.0 OR CC0-1.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -16,4 +16,38 @@ opt-level = 1
...
+bevy = "0.14"
+# Required for web builds.
+wasm-bindgen = { version = "0.2", optional = true }
+
+[features]
+default = [
+    # Default to a native dev build.
+    "native_dev",
+]
+core = []
```
MiniaczQ, line 28: Add a comment for what sort of features go here

MiniaczQ (changes requested):

Cargo.toml
benfrankel: Don't need this feature in retrospect, removed.

MiniaczQ (approved):

Cargo.toml
```diff
@@ -16,4 +16,35 @@ opt-level = 1
 opt-level = 3
 
 [dependencies]
-bevy = "0.12"
+bevy = "0.14"
+# Required for web builds.
+wasm-bindgen = { version = "0.2", optional = true }
```
janhohenheim:
```suggestion

[target.'cfg(target_family = "wasm")'.dependencies]
wasm-bindgen = "0.2"
```

Cargo.toml
```diff
@@ -16,4 +16,35 @@ opt-level = 1
...
+
+[features]
```
janhohenheim, line 24: Trim these to only `dev` and `dev_native`.

janhohenheim (changes requested):

janhohenheim (approved):

Issue #18, opened 2024-07-04
Add a `dev` feature or similar instead of `debug_assertions`


benfrankel:
A few templates do this. Not 100% sure why. Maybe to allow dev tools to be compiled in / out for both debug and release builds, e.g. including web builds.

The feature can be named `dev`, `debug`, or something else. It should be a default feature so you can just do `cargo run`.

benfrankel:
If this is added, existing references to `debug_assertions` should be updated.

janhohenheim:
Reason to have this:
```toml
[features]
default = ["native_dev"]
native_dev = ["bevy/dynamic_linking", "dev"]
wasm_dev = ["dev"]
dev = [ "some", "other", "features" ]
```

PR #24, opened 2024-07-05
Optimize dev experience for different targets

janhohenheim:
Alternative to #20 and #18
I'd prefer to keep a `dev` feature out until we can't avoid it in order to make it as simple as possible to use the template without fiddling with features.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /target
+/wasm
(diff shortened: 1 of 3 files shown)


janhohenheim:
Oh, `dev_dependencies` is only used for examples, tests, and benchmarks. Whoops.

PR #26, opened 2024-07-05
Add fast config

### LINKED ISSUE #13: Add `fast_builds_config.toml`
Leave out the part at the end pertaining to ci when copying from the bevy repo

janhohenheim:
Resolves #13

### DIFF
--- a/.cargo/config_fast_builds.toml
+++ b/.cargo/config_fast_builds.toml
@@ -1,0 +1,147 @@
+# Copy this file to `config.toml` to speed up your builds.
+#
+# # Faster linker
+#
+# One of the slowest aspects of compiling large Rust programs is the linking time. This file configures an
+# alternate linker that may improve build times. When choosing a new linker, you have two options:
+#
+# ## LLD
+#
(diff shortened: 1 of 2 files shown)


benfrankel:
The file should be in `.cargo/config.toml`, not `.cargo/config_fast_builds.toml`, I believe?

benfrankel (approved):

.cargo/config_fast_builds.toml
```diff
@@ -0,0 +1,147 @@
...
+# LLD linker
+#
+# You may need to install it:
+#
+# ```
+# cargo install -f cargo-binutils
+# rustup component add llvm-tools
+# ```
+linker = "rust-lld.exe"
+rustdocflags = ["-Clinker=rust-lld.exe"]
```
MiniaczQ, line 137: Let's keep it consistent, either all changes are commented out by default or none.

Should be fine with uncommenting all LLD code, since this is a patch and should do something when copy pasted.
So this would mean uncommenting LLD for other platforms

MiniaczQ (approved):

PR #27, opened 2024-07-05
Add .editorconfig

janhohenheim:


benfrankel (approved):
These style options seem universal / rustfmt defaults, so fair enough.

MiniaczQ (approved):

PR #29, opened 2024-07-05
Enable dynamic linking

janhohenheim:
I’m usually on Windows and every time dynamic linking was an issue, it was fixable by fiddling with the config, which is something we can make sure works out of the box in a template. 
Would you be okay with leaving it in until anyone reports an issue with it? I'd really like to have it be opt-out and not opt-in because this setting is by far the biggest thing you can do to improve your compile times.


MiniaczQ:
Where's the fiddling with the config?

janhohenheim:
@MiniaczQ not needed in this case, as it's a simple setup without LLD, cranelift, etc. 
The opt-levels should be enough.

MiniaczQ (approved):
Made sure it runs fine

PR #39, opened 2024-07-06
Fix and improve `release.yaml` workflow

benfrankel:
Relevant to https://github.com/TheBevyFlock/bevy-template/issues/19.

Overview of changes:
- Trigger release on pushed tag of the format `vA.B.C` instead of _any_ tag.
- Fix "prepare package" step of web release job (copy from the `wasm-release` folder instead of `release`).
- Set correct project and itch.io names.
- Add `wasm-opt` step for web release.
- Add `name` field to every step.
- Pull duplicate values into local `env` variables.
- Add helpful comments.
- Clean up formatting, names, wording, etc.


.github/workflows/release.yaml
```diff
@@ -1,280 +1,358 @@
 name: Release
 
+# Only trigger this workflow when a tag is pushed in the format `vA.B.C`.
 on:
   push:
     tags:
-      - '*'
-
+      - "v[0-9]+.[0-9]+.[0-9]+*"
```
BD103, line 7: Do the dots have to be escaped, since it's Regex?

```suggestion
      - "v[0-9]+\.[0-9]+\.[0-9]+*"
```

.github/workflows/release.yaml
```diff
@@ -1,280 +1,358 @@
...
+
+# Configure constants for this workflow.
 env:
-  # update with the name of the main binary
-  binary: bevy_github_ci_template
-  add_binaries_to_github_release: true
-  #itch_target: <itch.io-username>/<game-name>
-
+  # The base filename of the executable produced by `cargo build`.
+  EXE_NAME: bevy_template
```
BD103, line 18: I prefer the term binary instead of EXE because it's cross-platform.

.github/workflows/release.yaml
```diff
@@ -1,280 +1,358 @@
...
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
         with:
-          targets: x86_64-unknown-linux-gnu
-      - name: install dependencies
+          targets: ${{ env.TARGET }}
+
+      - name: Install dependencies
         run: |
           sudo apt-get update; sudo apt-get install pkg-config libx11-dev libasound2-dev libudev-dev libxcb-render0-dev libxcb-shape0-dev libxcb-xfixes0-dev
```
BD103, line 128: A few of these dependencies can be removed; I believe you only need the 4 listed [here](https://github.com/bevyengine/bevy/blob/09d86bfb96ccb66020c38485647c002dcfa37956/.github/actions/install-linux-deps/action.yml#L45).

.github/workflows/release.yaml
```diff
@@ -1,280 +1,358 @@
...
 
-      - name: Package as a zip
-        working-directory: ./linux
+      - name: Compress package
         run: |
-          zip --recurse-paths ../${{ env.binary }}.zip .
-
-      - name: Upload binaries to artifacts
+          zip --recurse-paths ${{ env.PACKAGE_NAME }}-${{ env.PLATFORM }}.zip ${{ env.OUT_DIR }}
+
```
BD103, line 143:
```suggestion

```

Just a bit of leftover whitespace :)

BD103 (approved):
Looks good! I have a few notes, but nothing major. I think in the future this can be converted to a matrix due to the amount of duplicated code, but that's probably for a follow-up PR :)

.github/workflows/release.yaml
MiniaczQ: Yes, the dot matches dot characters so this works but is overscoped

.github/workflows/release.yaml
benfrankel: I suppose that makes sense. `EXE` is meant to be short for `EXECUTABLE` here, not `.exe`, but I see that that's ambiguous.

benfrankel:
> Jan mentioned earlier that if we modify the base CI template we should submit the changes to the upstream, I think I agree with that

Yeah that's probably true. Note that their `env` variables and `cargo build` commands will look a bit different.

.github/workflows/release.yaml
benfrankel:
```suggestion
          sudo apt-get update; sudo apt-get install libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
```

.github/workflows/release.yaml
benfrankel: Done

.github/workflows/release.yaml
```diff
@@ -1,280 +1,357 @@
 name: Release
 
+# Only trigger this workflow when a tag is pushed in the format `vA.B.C`.
 on:
   push:
     tags:
-      - '*'
-
+      - "v[0-9]+\.[0-9]+\.[0-9]+*"
```
benfrankel, line 7:
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+*"
```

.github/workflows/release.yaml
benfrankel: So it turns out that Github workflows does not use regex: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet.

`.` character has no special meaning, it's just literal `.`

MiniaczQ (approved):
Looks good, best way to test it would be to run it 😉

PR #50, opened 2024-07-07
Move lints to `Cargo.toml` and add comments

benfrankel:


benfrankel:
Screwed something up with git. Will be easier to make a new PR.

PR #52, opened 2024-07-07
Add `.vscode/settings.json`

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy-template/issues/33.


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    // Allow `rust-analyzer` and `cargo` to run simultaneously.
+    // This uses a heavy amount of storage space, so consider commenting it out.
```
janhohenheim, line 3:
```suggestion
    // This uses extra storage space, so consider commenting it out.
```
It's not that heavy, as rust-analyzer doesn't do full builds.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
...
+    "rust-analyzer.cargo.targetDir": true
+}
```
janhohenheim, line 5: Do you have the editorconfig plugin enabled? There should be a terminating newline inserted automatically here

janhohenheim (changes requested):

.vscode/settings.json
benfrankel: Nope! Installing.

.vscode/settings.json
benfrankel: I'll take your word for it, I haven't tried this setting myself yet.

MiniaczQ (approved):
Works as intended

janhohenheim (approved):

PR #54, opened 2024-07-07
Split stuff into lib and main

janhohenheim:


benfrankel (changes requested):
We'll probably still want `src/core/`. The split becomes more obvious when there's more cross-cutting stuff and more game-logic stuff later on. It can be re-added at that point so I'm not too concerned about it for now, though.

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 .cargo/config.toml
+/wasm
```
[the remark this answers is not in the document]
janhohenheim: Running local Wasm builds. I added it because I happened to be running wasm before, so these folders were laying around.

src/lib.rs
```diff
@@ -0,0 +1,38 @@
+// Disable console on Windows for non-dev builds.
+#![cfg_attr(not(feature = "dev"), windows_subsystem = "windows")]
```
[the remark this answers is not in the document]
janhohenheim: good catch!

MiniaczQ (approved):

.gitignore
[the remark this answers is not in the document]
benfrankel: Makes sense. TBH some comments in the `.gitignore` may be reasonable (for the other 2 lines as well), but can be a separate PR.

src/lib.rs
```diff
@@ -0,0 +1,33 @@
+mod bevy_setup;
+mod core;
+#[cfg(feature = "dev")]
+mod dev;
+mod ui;
+
+use bevy::prelude::*;
+
+pub struct GamePlugin;
```
benfrankel, line 9: Wondering if this should be `pub fn plugin` for consistency, and `.add_plugins(bevy_template::plugin)` in `main.rs`.

src/bevy_setup/mod.rs
```diff
@@ -0,0 +1,32 @@
+use bevy::asset::AssetMetaCheck;
```
benfrankel, line 1: Any reason this file is `bevy_setup/mod.rs` instead of `bevy_setup.rs`?

src/core/loading/mod.rs
```diff
@@ -1,14 +1,13 @@
+use crate::AppState;
```
benfrankel, line 1: Crate-local imports should come after 3rd-party imports, with a blank line separating, like before.

This comment applies to the other places this was changed as well.

src/core/loading/mod.rs
```diff
@@ -1,14 +1,13 @@
...
-use super::AppState;
+use bevy::prelude::*;
 
 pub(super) fn plugin(app: &mut App) {
     // Setup(s), update(s), teardown(s)
     app.add_systems(Update, frames_elapsed.run_if(in_state(AppState::Loading)));
 }
 
-// Based on: <https://github.com/bevyengine/bevy/blob/v0.14.0/examples/window/window_settings.rs#L56>
+/// Based on: <https://github.com/bevyengine/bevy/blob/v0.14.0/examples/window/window_settings.rs#L56>
```
benfrankel, line 10: Doc comment for private function?

benfrankel (changes requested):
It seems we're structuring modules in `core/` based on states? I'd argue this is an anti-pattern. Also `dev.rs` should go in `core/`.

src/bevy_setup/mod.rs
janhohenheim: Consistency with the others. But you're right, I'll change it.

src/core/loading/mod.rs
janhohenheim: Yes, for IDE support.

src/core/loading/mod.rs
janhohenheim: I'll see if I can just add a `.rustfmt` for that

src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+        app.init_state::<AppState>()
+            // Add state scoped entities for UI cleanup
+            .enable_state_scoped_entities::<AppState>()
+            // Sub-plugins
+            .add_plugins((bevy_setup::plugin, core::plugin, ui::plugin));
+
+        #[cfg(feature = "dev")]
+        app.add_plugins(dev::plugin);
+
+        // For a larger UI example visit: https://github.com/MiniaczQ/bevy-substate-project
```
MiniaczQ, line 23: Remove for now, we will add it with the rest of opinionated templates

src/lib.rs
MiniaczQ: I think it's special enough to be this way

benfrankel: Understandable.

src/core/loading/mod.rs
janhohenheim: Update: yes, `rustfmt.toml` has an option for that, but it's nightly only. I've applied it but have not added it to the repo because it will generate warnings on stable.

benfrankel: My `rustfmt.toml`:
```toml
# Stable
max_width = 100
use_field_init_shorthand = true
use_try_shorthand = true

# Unstable
group_imports = "StdExternalCrate"
imports_granularity = "Item"
imports_layout = "HorizontalVertical"
match_block_trailing_comma = true
normalize_doc_attributes = true
```
This is my personal preference, not a suggestion for this repo, necessarily.

src/core/loading/mod.rs
benfrankel: Huh ok.

janhohenheim: I like my mouseovers to show pretty docs ;)

benfrankel (approved):
Approving to avoid infinite code review loop and move forward with more PRs :)

MiniaczQ (approved):

PR #55, opened 2024-07-07
Move transition logging to dev 

janhohenheim:
Depends on #54


MiniaczQ (approved):

benfrankel (approved):
Obligatory pyri_state handles this better :)

PR #56, opened 2024-07-07
Refactor project structure

benfrankel:
Verified that this compiles and seems to still run correctly.

Overview of changes:
- Split `bevy_setup.rs` into `core/window.rs` and `core/asset.rs`.
- Moved camera setup out of `core/mod.rs` and into `core/camera.rs`.
- Moved "main app state" stuff out of `core/` and into `screen/`.
- Flattened `AppState`, `RunningState`, and `MenuState` into `Screen`.
- Renamed `Loading` state to `Boot` to make it clear that it's at app startup, and there may be another loading screen for assets later on.
- Moved "make window invisible and then later visible again" logic into `screen/boot.rs`.
- Renamed `Menu` state to `Title` to be more explicit (there may be multiple menus in a game, but only one title screen).
- Renamed systems to be more explicit.
- Renamed `UiAction` components to be more explicit.
- Added `src/game/mod.rs` for game mechanics / logic (currently empty).
- Moved `ui/` into the new `util/` folder which allows certain lints.


src/screen/boot.rs
```diff
@@ -0,0 +1,36 @@
...
+        Update,
+        continue_from_boot_after_delay.run_if(in_state(Screen::Boot)),
+    );
+}
+
+fn enter_boot(mut window: Query<&mut Window>) {
+    // This will make the window invisible during `Boot`.
+    // This is useful when you want to avoid the white flash that shows up before
+    // the GPU is ready to render the app. Based on: <https://github.com/bevyengine/bevy/blob/v0.14.0/examples/window/window_settings.rs#L56>
+    // This workaround is not needed on Windows: <https://github.com/bevyengine/bevy/issues/14135>
```
janhohenheim, line 20:
```suggestion
    // This workaround does not currently work on Windows: <https://github.com/bevyengine/bevy/issues/14135>
```

src/screen/playing.rs
```diff
@@ -0,0 +1,32 @@
...
+        SpriteBundle {
+            texture: asset_server.load("ducky.png"),
+            ..Default::default()
+        },
+        StateScoped(Screen::Playing),
+    ));
+}
+
+// TODO: Reset camera transform here.
+fn exit_playing() {}
```
janhohenheim, line 28: What do you mean by this?

src/util/mod.rs
```diff
@@ -0,0 +1,12 @@
+//! Self-contained, re-usable utilities that are not specific to this game.
+
+#![allow(dead_code)]
+#![allow(unused_imports)]
```
janhohenheim, line 4: Don't think this one's needed?

src/core/mod.rs
```diff
@@ -1,46 +1,28 @@
...
-        OnEnter(AppState::Ready),
-        (make_window_visible, setup_camera),
+    // Bevy plugins.
+    app.add_plugins(
+        DefaultPlugins
+            .build()
+            .disable::<AssetPlugin>()
+            .add_after::<AssetPlugin, _>(asset::plugin)
+            .disable::<WindowPlugin>()
+            .add_after::<WindowPlugin, _>(window::plugin),
```
MiniaczQ, line 19: Document this

src/screen/playing.rs
benfrankel: If the camera moves around while you're playing (e.g. follow the player), which may not be universal but happens in almost all games, its transform should be reset to 0, 0 when you exit the playing state.

src/screen/boot.rs
benfrankel: Ah I see.

src/screen/boot.rs
```diff
@@ -0,0 +1,36 @@
...
+
+/// Based on: <https://github.com/bevyengine/bevy/blob/v0.14.0/examples/window/window_settings.rs#L56>
+fn continue_from_boot_after_delay(
+    mut next_screen: ResMut<NextState<Screen>>,
+    frames: Res<FrameCount>,
+) {
+    if frames.0 >= 3 {
+        next_screen.set(Screen::Title);
+    }
+}
```
MiniaczQ: Now that you have it all in one place it deserves it's own `deflicker` plugin in `core`
Then the "Based on" can be the module/plugin's documentation

src/util/mod.rs
benfrankel: Well I can remove these lints for now. IME these warnings tend to come up undesirably in `util/` code, but there won't be a warning if I remove them for this PR.

src/screen/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+#[derive(States, Debug, Hash, PartialEq, Eq, Clone, Default)]
+pub enum Screen {
+    #[default]
+    Boot,
+    Title,
+    Credits,
+    Playing,
+}
```
MiniaczQ, line 27: Not a big fan of flattening this,
I'd at least separate `Boot` out of here into a parent state, since you very clearly aren't supposed to go back there.
Including at least one hierarchy here would also help to show how this scales.

src/util/mod.rs
janhohenheim: Definitely agree for `dead_code`, but I'm curious about how `unused_imports` comes up.

benfrankel: That's from re-exports.

src/screen/title.rs
```diff
@@ -0,0 +1,47 @@
...
+        .my_button("Play")
+        .insert(TitleAction::Play)
+        .set_parent(list);
+    commands
+        .my_button("Credits")
+        .insert(TitleAction::Credits)
+        .set_parent(list);
+}
+
+fn exit_title() {}
```
MiniaczQ, line 33: 🙃

src/screen/boot.rs
benfrankel: Fair enough, will be a separate PR.

src/screen/mod.rs
benfrankel: Fair enough, will be a separate PR.

src/screen/title.rs
benfrankel: This is just for consistency but I can definitely remove it. State-scoped entities making exit systems redundant :weary:

src/core/mod.rs
benfrankel: Done.

src/util/mod.rs
benfrankel: Added this back in, commented out and explained.

src/util/mod.rs
```diff
@@ -0,0 +1,14 @@
+//! Self-contained, re-usable utilities that are not specific to this game.
+
+// Unused utilities and re-exports may trigger these lints undesirably.
+// Consider uncommenting these lines.
+// #![allow(dead_code)]
+// #![allow(unused_imports)]
```
janhohenheim, line 6:
```suggestion
#![allow(dead_code, unused_imports)]
```
Would be a better default dev experience imo

janhohenheim (changes requested):

src/screen/title.rs
benfrankel: Done.

src/util/mod.rs
benfrankel: Sure.

src/screen/title.rs
MiniaczQ: It's good for clearing resources which cannot be scoped yet

janhohenheim (approved):

src/screen/title.rs
MiniaczQ: but yeah, adding that empty function is pedantic 😄

MiniaczQ (approved):

PR #59, opened 2024-07-08
Do some minor refactoring

janhohenheim:
- Changed visibility modifiers. This is not important for a game, but it makes unused code detection more accurate
- Added some comments
- Removed some comments that I thought didn't add much
- Chained `app.foo` calls

### DIFF
--- a/src/screen/mod.rs
+++ b/src/screen/mod.rs
@@ -1,4 +1,4 @@
-//! The states describing the game's main screen states and the transitions between them.
+//! The game's main screen states and transitions between them.
 
 mod credits;
 mod playing;
--- a/src/screen/mod.rs
+++ b/src/screen/mod.rs
@@ -1,3 +1,5 @@
+//! The states describing the game's main screen states and the transitions between them.
+
 mod credits;
 mod playing;
 mod title;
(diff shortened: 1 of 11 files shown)


src/screen/mod.rs
```diff
@@ -1,3 +1,5 @@
+//! The states describing the game's main screen states and the transitions between them.
```
benfrankel, line 1:
```suggestion
//! The game's main screen states and transitions between them.
```

src/screen/mod.rs
```diff
@@ -8,14 +10,14 @@ use crate::core::booting::Booting;
 
 pub(super) fn plugin(app: &mut App) {
     app.init_state::<Screen>()
-        .enable_state_scoped_entities::<Screen>();
-
-    app.add_plugins((title::plugin, credits::plugin, playing::plugin));
+        .enable_state_scoped_entities::<Screen>()
+        .add_plugins((title::plugin, credits::plugin, playing::plugin));
```
benfrankel, line 14: I like splitting the `app.` chains based on logical grouping. The first two methods are for setting up `Screen`, and the third method is for adding sub-plugins. This plugin is small enough that putting it all in one chain is still readable, but for larger plugins it may hurt readability.

Willing to compromise on this.

benfrankel (changes requested):

src/screen/mod.rs
janhohenheim: Agreed on grouping in bigger plugins, although I would just place a comment above a logical group and still chain it.
But in this case, it should be all clear enough that the terse version makes it more lightweight.

Also willing to compromise, not a strong opinion.

MiniaczQ: 😈

janhohenheim: [`8666105` (#59)](https://github.com/TheBevyFlock/bevy-template/pull/59/commits/86661056d0ef97f89c520aa75b9e60ef5ce9063f)

benfrankel: I'm convinced by the mini argument. Never using `app.` chains anywhere is superior.

benfrankel (approved):

PR #66, opened 2024-07-08
Add splash screen

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy-template/issues/58.

Minimal implementation. Features the Bevy logo from here: https://github.com/bevyengine/bevy/blob/main/assets/branding/bevy_logo_dark.png


janhohenheim (commented):
Code looks good, I'll approve once I see how it looks while running it on my machine

benfrankel:
Expect it to be pretty jarring with no fade in / out :P

janhohenheim (changes requested):
It's blue for a split-second before the splash screen appears on Windows, but I think that's because of the missing deflicker hack. Still, it might be good to have the `ClearColor` match the splash to be sure.

benfrankel:
> It's blue for a split-second before the splash screen appears on Windows, but I think that's because of the missing deflicker hack. Still, it might be good to have the `ClearColor` match the splash to be sure.

Added `ClearColor` in `screen/splash.rs` with a comment explaining it, since it's a workaround for a bug.

janhohenheim (approved):
Perfect, thanks!

MiniaczQ (approved):

PR #67, opened 2024-07-08
Refactor `util/ui/` and add prelude

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy-template/issues/64.

This also removes `ClearColor` and instead uses the `BackgroundColor` component of `ui_root`.


benfrankel:
Rationale for removing `With<Button>`: Not every interactable entity will be a proper `Button`. It may be a hyperlink, aka text, or something else.

janhohenheim (approved):
Makes sense to me :)

PR #70, opened 2024-07-08
Add player movement code

### LINKED ISSUE #10: Setup a basic `SystemSet` enum for the game
*Migrated from https://github.com/bevyengine/bevy_github_ci_template/issues/63*

Same rationale as #4

### LINKED ISSUE #11: Add a basic game interaction
Something that paints a metaphorical "YOUR GAME LOGIC HERE" sign.
I propose having a simple 2D sprite that moves on WASD like in https://github.com/NiklasEi/bevy_game_template

janhohenheim:
Adapted from https://github.com/bevyengine/bevy/pull/14223
Resolves #10 
Resolves #11 

I'm intentionally using `FixedUpdate` here instead of a very minimal example since I expect many jam submissions to do some kind of player movement, so it's best to show them how it's done correctly right away.
I also moved the level spawning away from screen because that's a game detail and not a screen transition detail.

The input is not nicely abstracted away because when I started doing that, I noticed I was basically reimplementing LWIM. Better to just link to that imo.

I know that @benfrankel said that physics go in `core`, but I intentionally placed them in `game` here because I expect the physics handling to be vastly different from game to game and I want to make it obvious to users that they are encouraged to change the implementation according to their needs. The rest of `core` is comparatively tame; I expect many will leave most of it untouched.


src/game/physics.rs
```diff
@@ -0,0 +1,51 @@
...
+    const STORAGE_TYPE: StorageType = StorageType::Table;
+
+    fn register_component_hooks(hooks: &mut ComponentHooks) {
+        hooks.on_add(|mut world, entity, _component_id| {
+            let rendered_transform = *world.get::<Transform>(entity).unwrap();
+            let mut physical_transform = world.get_mut::<PhysicalTransform>(entity).unwrap();
+            physical_transform.0 = rendered_transform;
+        });
+    }
+}
```
MiniaczQ, line 37: This is basically the entity version of `FromWorld`, maybe `FromWorldEntity` trait is something that we might want in Bevy

src/game/physics.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// How many units per second the player should move.
+#[derive(Debug, Component, Clone, Copy, PartialEq, Default, Deref, DerefMut)]
+pub(crate) struct Velocity(pub(crate) Vec3);
+
+/// The actual transform of the player in the physics simulation.
+/// This is separate from the `Transform`, which is merely a visual
+/// representation.
+
```
MiniaczQ, line 21:
```suggestion
```

src/game/level.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::screen::Screen;
+
+#[derive(Debug, Event)]
+pub(crate) struct SpawnLevel;
+
+pub(super) fn plugin(app: &mut App) {
+    app.observe(spawn_level);
+}
+
+fn spawn_level(
```
MiniaczQ, line 15: Maybe level -> player?
There is no level really

src/game/level.rs
```diff
@@ -0,0 +1,31 @@
...
+    _trigger: Trigger<SpawnLevel>,
+    mut commands: Commands,
+    asset_server: Res<AssetServer>,
+) {
+    commands.spawn((
+        Name::new("Player"),
+        SpriteBundle {
+            texture: asset_server.load("ducky.png"),
```
MiniaczQ, line 23: Do we want to add a loading stage for game assets later?

janhohenheim: Yep, that would be nice. #71

src/game/level.rs
janhohenheim: I'd leave it as-is so that users know where to put their level code.

src/game/physics.rs
janhohenheim: That would have a weird overlap with lifecycle hooks now though. Except maybe as a convenience trait that implemented hooks for you in the background?

src/game/level.rs
```diff
@@ -0,0 +1,31 @@
...
+            transform: Transform::from_scale(Vec3::splat(0.4)),
+            ..Default::default()
+        },
+        Velocity::default(),
+        PhysicalTransform::default(),
```
MiniaczQ, line 28: This works here, but physics engines usually rely on `Transform` directly, would the solution be as following?
- Root entity -- has child -> Physical entity
- Root entity -- has child -> Graphical entity
- FixedUpdate -- update_physics()
- Update -- interpolate_graphics()

I don't think we should add this needless boilerplate here, but it's more than worth few lines of comments, since basically every 2D game, top down or platformer uses some sorts of physics

janhohenheim: Good point, I added a comment.
I did not rename `update_rendered_transform` to `interpolate_graphics` because the fact that it uses interpolation instead of advanced simulation or extrapolation is a detail.

MiniaczQ: Yeah, the names aren't important at all here :)

src/game/level.rs
MiniaczQ: Discussed on discord https://discord.com/channels/691052431525675048/1258521739395203174/1259897153506115725

src/game/physics.rs
MiniaczQ: /shrug just an idea, I feel like we'll be seeing this pattern a lot

src/game/level.rs
janhohenheim: Per discussion on Discord, I changed it: [`cd4934f` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/cd4934f0536b731875d1150e1bc83b9691130b8a)

src/game/physics.rs
janhohenheim: Yeah, Avian physics already has it in place in a bunch of types. Let's see what ergonomics we develop in the future, hooks are still brand new after all :)

MiniaczQ: https://discord.com/channels/691052431525675048/749335865876021248/1259901488428224554

src/game/movement.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub(super) fn plugin(app: &mut App) {
+    app.add_systems(Update, handle_input);
+}
+
+/// Handle keyboard input to move the player.
+fn handle_input(keyboard_input: Res<ButtonInput<KeyCode>>, mut query: Query<&mut Velocity>) {
+    /// Since Bevy's default 2D camera setup is scaled such that
+    /// one unit is one pixel, you can think of this as
+    /// "How many pixels per second should the player move?"
```
MiniaczQ, line 15:
```suggestion
    /// "How many pixels per second should the player move?"
    /// Note that physics engines may use different unit/pixel ratios.
```

src/game/spawn/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Handles spawning of entities. This is typically done through
+//! [observers](https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.Observer.html).
```
MiniaczQ, line 2:
```suggestion
//! [observers](https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.Observer.html).
// Alternatively this can be done through `Events<E>` or `Commands`.
```

MiniaczQ (approved):

src/game/spawn/level.rs
```diff
@@ -0,0 +1,18 @@
+//! Spawn our main level by triggering other spawn events.
```
benfrankel, line 1:
```suggestion
//! Spawn the main level.
```

src/game/spawn/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! Handles spawning of entities. This is typically done through
```
benfrankel, line 1: "typically" according to whom? Observers were only just added :P

benfrankel (changes requested):

src/game/spawn/level.rs
janhohenheim: I want to make it explicit that users should not hardcode spawning logic in there and instead defer to other observers.

src/game/spawn/mod.rs
janhohenheim: According to us. We are the authority here :P

janhohenheim: [`7945346` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/79453462dd6fabb65ccad4203f0e2d2938147e0c)

benfrankel: I mean that this makes it sound like it's an old tradition in bevy.

src/game/spawn/player.rs
```diff
@@ -1,19 +1,21 @@
...
-use super::physics::{PhysicalTransform, Velocity};
-use crate::screen::Screen;
+use crate::{
+    game::physics::{PhysicalTransform, Velocity},
+    screen::Screen,
+};
 
 #[derive(Debug, Event)]
-pub(crate) struct SpawnLevel;
+pub(crate) struct SpawnPlayer;
```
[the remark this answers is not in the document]
janhohenheim: [`bfd64af` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/bfd64afbc6f19a06df0ddc7803b508d238db877e)

src/game/spawn/level.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use bevy::prelude::*;
+
+use super::player::SpawnPlayer;
+
+#[derive(Debug, Event)]
+pub(crate) struct SpawnLevel;
```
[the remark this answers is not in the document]
janhohenheim: [`bfd64af` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/bfd64afbc6f19a06df0ddc7803b508d238db877e)

src/game/spawn/level.rs
janhohenheim: Changed the phrasing a bit [`d1b5de8` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/d1b5de82b563baf6d240307f3fbdc22e2e115238)

src/game/physics.rs
```diff
@@ -0,0 +1,50 @@
...
+/// How many units per second the player should move.
+#[derive(Debug, Component, Clone, Copy, PartialEq, Default, Deref, DerefMut)]
+pub(crate) struct Velocity(pub(crate) Vec3);
+
+/// The actual transform of the player in the physics simulation.
+/// This is separate from the `Transform`, which is merely a visual
+/// representation.
+#[derive(Debug, Clone, Copy, PartialEq, Default, Deref, DerefMut)]
+pub(crate) struct PhysicalTransform(pub(crate) Transform);
```
benfrankel, line 22: I'm willing to believe this is needed but I'd really need to be convinced of it.

src/game/physics.rs
```diff
@@ -0,0 +1,50 @@
...
+            let mut physical_transform = world.get_mut::<PhysicalTransform>(entity).unwrap();
+            physical_transform.0 = rendered_transform;
+        });
+    }
+}
+
+/// Advance the physics simulation by one fixed timestep. This may run zero or
+/// multiple times per frame.
+///
+/// Note that since this runs in `FixedUpdate`, `Res<Time>` would be
```
benfrankel, line 41: I feel like being explicit within the system that its timestep will be fixed is not idiomatic. Same opinion for other instances of this in the PR ofc.

src/game/spawn/player.rs
```diff
@@ -0,0 +1,36 @@
...
+            texture: asset_server.load("ducky.png"),
+            transform: Transform::from_scale(Vec3::splat(0.4)),
+            ..Default::default()
+        },
+        Velocity::default(),
+        // If your physics engine of choice uses `Transform` directly,
+        // a good hierarchy to follow instead is to have a `Player` root entity
+        // with the physical transform and the rendered transform as individual children.
+        PhysicalTransform::default(),
+        StateScoped(Screen::Playing),
```
benfrankel, line 34: Ideally this would not be coupled. What if you want to like.. spawn a lil player entity in the title screen as a fun lil thing?

Fine to potentially change this in a later PR though.

benfrankel (changes requested):
Comments on the code, not on the physics mainly.

src/game/spawn/level.rs
benfrankel: Okay, fair enough.

src/game/movement.rs
```diff
@@ -0,0 +1,38 @@
...
+use bevy::prelude::*;
+
+use super::physics::Velocity;
+
+pub(super) fn plugin(app: &mut App) {
+    app.add_systems(Update, handle_input);
+}
+
+/// Handle keyboard input to move the player.
+fn handle_input(keyboard_input: Res<ButtonInput<KeyCode>>, mut query: Query<&mut Velocity>) {
```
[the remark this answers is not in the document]
benfrankel: Oh also. `handle_input` is an ambiguous name. Consider something like `record_movement_input`.

src/game/physics.rs
janhohenheim: I disagree on that, as the user might not be familiar with how time steps work, so they might not realize that this `Time` is different from the `Time` they're used to. @MiniaczQ, can you do a tie-breaker? I'll accept it if that's the majority opinion.

src/game/physics.rs
MiniaczQ: This is needed because we cannot store 3 Transforms on a single entity without newtyping

src/game/physics.rs
MiniaczQ: Imo systems should be designed with schedule in mind, I'm not a huge fan of `Time<()>`

src/game/physics.rs
benfrankel: Well yeah, the question is moreso why do we need another Transform :smile:

src/game/spawn/player.rs
MiniaczQ: `Box<dyn Bundle>`, this will allow for state scoping as well as naming

benfrankel: That might be fine. Hopefully bsn will take the wheel soon enough.

src/game/physics.rs
MiniaczQ: `Transform` is used for rendering, we need to store another one for FixedUpdate as well as a buffer for it's previous value

src/game/spawn/player.rs
janhohenheim: > `Box<dyn Bundle>`, this will allow for state scoping as well as naming and any sort of marker components

Could you elaborate on that? I'm curious.

benfrankel: `commands.spawn` uses `Commands::entities` to return an `EntityCommands` which includes both `Entity` and `Commands`.

`commands.add` and `commands.trigger` both can't do this, but an extension trait could (like we do for UI widgets). Maybe an extension trait with a generic method like `commands.my_spawn(impl Spawnable) -> EntityCommands`.

benfrankel: Gonna resolve this conversation because it can be discussed further on discord and handled in a later PR.

janhohenheim: Leaving this for a follow-up

src/game/physics.rs
janhohenheim: See https://github.com/bevyengine/bevy/pull/14223 for a full writeup.

src/game/physics.rs
janhohenheim: [`a4c1565` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/a4c1565bd7d5e0fdb41255e19ea148d77c0c24b4)

src/game/movement.rs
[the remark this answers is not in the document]
janhohenheim: [`7ae0325` (#70)](https://github.com/TheBevyFlock/bevy-template/pull/70/commits/7ae0325c9e01159448faceec4ff5ac5c216c8341)

src/game/spawn/player.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::{
+    game::physics::{PhysicalTransform, Velocity},
+    screen::Screen,
+};
+
+pub(super) fn plugin(app: &mut App) {
+    app.observe(spawn_player);
+}
+
+#[derive(Debug, Event)]
```
benfrankel, line 14: This and other derives should put `Event`, or `Component`, etc. first.

benfrankel (approved):
Not blocking on comments, better to get this in now so it can be iterated on.

PR #75, opened 2024-07-08
Refactor deferred spawns

MiniaczQ:


src/util/spawn.rs
```diff
@@ -0,0 +1,30 @@
+use bevy::{ecs::system::EntityCommands, prelude::*};
+
+/// Things that can spawn entities.
```
benfrankel, line 3:
```suggestion
/// Types that can spawn entities.
```

MiniaczQ:
I honestly think this is taking it too far...
The PoC works, but the API is pretty horrendous.

src/util/spawn_trigger.rs
benfrankel, line 1: Would move `spawn.rs` to `spawn/mod.rs` and this to `spawn/trigger.rs`.

benfrankel (changes requested):
Preliminary comments, will review more.

src/util/spawn_trigger.rs
```diff
@@ -0,0 +1,21 @@
...
+use super::spawn::Spawn;
+
+/// Trait for spawning empty entitie and triggering them.
+pub(crate) trait SpawnTrigger {
+    /// Spawns an empty entity and triggers it with the event.
+    fn spawn_trigger<E: Event>(&mut self, event: E) -> EntityCommands;
+}
+
+impl<T: Spawn> SpawnTrigger for T {
+    fn spawn_trigger<E: Event>(&mut self, event: E) -> EntityCommands {
```
benfrankel, line 12: Maybe this would read better renamed to `TriggerSpawn::trigger_spawn`?

src/util/spawn_trigger.rs
```diff
@@ -0,0 +1,21 @@
...
+        // Type juggling to get around lifetime downcasting.
+        // We cannot recover `entity_commands` if we ever drop it.
+        let mut entity_commands = self.spawn_empty();
+        let entity = entity_commands.id();
+        let mut commands = entity_commands.commands();
+        commands.trigger_targets(event, entity);
+        entity_commands
```
benfrankel, line 19:
```suggestion
        let mut e = self.spawn_empty();
        let entity = e.id();
        e.commands().trigger_targets(event, entity);
        e
```

benfrankel: Just a lil smaller until we get `EntityCommands::trigger`

benfrankel (changes requested):

src/util/spawn_trigger.rs
MiniaczQ: But it spawns then triggers 😛

src/game/spawn/mod.rs
```diff
@@ -1,12 +1,30 @@
...
+    fn get_trigger_entity_or_spawn<E>(&mut self, trigger: Trigger<E>) -> EntityCommands {
+        let entity = trigger.entity();
+        // Cannot use `match` due to compile time trait restraints
+        if entity == Entity::PLACEHOLDER {
+            self.spawn_empty()
+        } else {
+            self.entity(entity)
+        }
+    }
+}
```
benfrankel, line 30: This can be deleted if we expect these events to always be triggered with an entity. I think that's valid, since we're providing `spawn_trigger` as the interface for triggering these events.

Then you can do `commands.entity(trigger.entity()).insert(...)` etc

benfrankel (changes requested):

src/util/spawn_trigger.rs
benfrankel: Yeah. Implementation detail though, can read it as "triggering a spawn", which is imo conceptually valid.

src/game/spawn/mod.rs
MiniaczQ: Spawn trigger is fully separate from this api, its not obvious

src/util/ui/widgets.rs
```diff
@@ -2,39 +2,23 @@
...
-
-impl<'a> Spawner<'a> for ChildBuilder<'a> {
-    fn spawn<B: Bundle>(&mut self, bundle: B) -> EntityCommands {
-        self.spawn(bundle)
-    }
-}
-
-/// Root container spawning trait.
-pub(crate) trait RootContainers {
+/// Contanier spawning trait.
```
janhohenheim, line 14:
```suggestion
/// Container spawning trait.
```

janhohenheim (changes requested):

MiniaczQ:
Closing this as agreed

PR #76, opened 2024-07-08
Support arrow keys for movement

benfrankel:


src/game/movement.rs
```diff
@@ -13,32 +13,33 @@ pub(super) fn plugin(app: &mut App) {
...
+    // Need to normalize and scale because otherwise
+    // diagonal movement would be faster than horizontal or vertical movement.
+    let intent = intent.normalize_or_zero() * SPEED;
 
-        // Need to normalize and scale because otherwise
-        // diagonal movement would be faster than horizontal or vertical movement.
-        velocity.0 = velocity.normalize_or_zero() * SPEED;
+    for mut velocity in &mut velocity_query {
+        velocity.0 = intent;
     }
```
MiniaczQ, line 44: I'd go as far as `get_single_mut` here

src/game/movement.rs
```diff
@@ -13,32 +13,31 @@ pub(super) fn plugin(app: &mut App) {
...
-            velocity.x -= 1.0;
-        }
-        if keyboard_input.pressed(KeyCode::KeyD) {
-            velocity.x += 1.0;
-        }
-
-        // Need to normalize and scale because otherwise
-        // diagonal movement would be faster than horizontal or vertical movement.
-        velocity.0 = velocity.normalize_or_zero() * SPEED;
+    let mut intent = Vec3::ZERO;
```
janhohenheim, line 25: Not a fan of this rename. @MiniaczQ, opinion?

janhohenheim (changes requested):

src/game/movement.rs
benfrankel: Rationale: Input -> Intent -> Behavior opens the door for EnemyAi -> Intent -> Behavior, so this is the name I would use for the intermediate value.

benfrankel: Although I suppose the intent wouldn't have speed applied.

MiniaczQ: I'm ok with `intent` as long as it's axis-like, after we multiply it with SPEED it's very clearly `velocity` and no longer `intent`.

janhohenheim (approved):

PR #79, opened 2024-07-08
Simplify

cart:
I have some proposed changes, in the interest of making this all a bit easier to grok. In general I think we've gone too far down the path of modularization and abstraction. I think this comes at the cost of making this look and feel "big" and "complex" when it should look and feel "small" and "simple". Note this isn't just about appearances / first impressions. I strongly believe that the new layout is better for developer experience generally.

1. I think "core" should be flattened out into the top level. GamePlugin (now renamed to AppPlugin to avoid the conceptual overlap with `game::plugin`) is the right place to be the "top level view" of the whole app.
2. Remove `util` plugin in favor of flattening out at the top based on "top level context"
3. Removed separate WindowPlugin / AssetPlugin init plugins. The "disable and add_after(new_window_plugin)"  approach is cumbersome and overly abstracted. I think it makes a lot more sense to do this at the top level. If we really want these declarations in separate files (imo this isn't worth it in a template), we should favor just putting them in a `window_plugin()` function and calling `DefaultPlugins.set(window_plugin())`.
4. Rename `util::ui` to top-level `ui_tools` module.
5. Merge screen plugin / boot plugin / etc into a single plugin. In effect, these are all related to each other / share the same purpose. The "boot" state is really only useful in the context of deflickering the screen anyway. I kind of think we should drop the "deflicker" fix (and the boot state) in favor of trying to resolve that upstream, as code like this existing in a template is a pretty bad look (and I don't think this is something most app developers should be thinking about). I'd almost prefer to completely remove it and just let things flicker in the interim. 
6. Favor `pub` instead of `pub(crate)` as this distinction seems unimportant and boilerplatey in this context.


cart:
Also: the template was panicking prior to this change ... I did not introduce that.

benfrankel:
Regarding 3, for window plugin at least, there are other things that would want to go in there like a window config asset (fullscreen, resolution, etc.) and `WindowState { Booting, Ready }` from deflicker, which would include the deflicker as well as loading the window config. Putting the bevy window plugin in the same place keeps all the windowing logic localized.

cart:
I would personally prefer 3rd party "extra" window logic to exist elsewhere, rather than "replacing" the window plugin. That is conceptually clearer (and also serves to reduce unnecessary boilerplate).

benfrankel:
It could provide a const `WindowPlugin` value instead of adding the plugin itself, I guess.

cart:
> It could provide a const WindowPlugin value instead of adding the plugin itself, I guess.

Consider just doing:

```rust
app.add_plugins(DefaultPlugins.set(WindowPlugin {
  primary_window: Some(primary_window),
  ..default()
});

// later in your window module
fn primary_window() -> Window {
    Window {
        title: "bevy-template".to_string(),
        canvas: Some("#bevy".to_string()),
        fit_canvas_to_parent: true,
        prevent_default_event_handling: true,
        ..default()
    }
}
```

cart:
>  loading and applying a window config asset (fullscreen, resolution, etc.) and WindowState { Booting, Ready } 

I personally think both of these things are likely overkill for a template, which should be focused on bare minimum engine config and app logic. If we want these features, imo we should focus on implementing them at the engine level, rather than trying to meet every possible need at the template level.

janhohenheim (approved):
Per feedback on Discord, this seems to mirror the general impression others had. Looking at the code, the changes seem pretty alright. I disagree in places because I have a certain idea of how the current structure would nicely adapt to future expansions, but I also see the need to keep things simple. In favor of merging this.

janhohenheim:
@cart 
> I personally think both of these things are likely overkill for a template, which should be focused on bare minimum engine config and app logic.

I don't have any strong opinions on fullscreen and resolution, but I disagree with the second half. There already exists such a template: https://github.com/bevyengine/bevy_github_ci_template
This one was borne out of the sentiment that said template was nice, but too bare-bones to be useful to many people entering a jam. As such, we try to intentionally add a bit more than just the bare minimum and actively show how a real game might be structured.

MiniaczQ (changes requested):
image
We need a dedicated UI camera or revert on the camera move

benfrankel:
From discord:

> regarding @cart's PR. i'd prefer if the changes were split into several PRs
> some of the changes are uncontroversial.. some are more controversial
> and i don't want to have to resolve all of the controversial stuff while the greater PR is looming the entire time, freezing other potential work and its uncontroversial parts 
> i'm in favor of simplifying, but not necessarily the way the PR does it
>
> i think a practical plan would be to start by taking the uncontroversial changes from cart's PR and reimplementing them as separate PRs, and merging those through
> then discuss the controversial bits
> and just use cart's PR as a reference instead of actually trying to merge it wholesale

benfrankel:
Closing as outdated. These changes have been applied (not necessarily verbatim) as a few smaller PRs.

PR #87, opened 2024-07-09
Flatten `src/core/` into `src/`

benfrankel:


src/lib.rs
```diff
@@ -1,14 +1,35 @@
...
+        // Add Bevy plugins.
+        app.add_plugins(
+            DefaultPlugins
+                .build()
+                // Add `AssetPlugin` via `asset::plugin`.
+                .disable::<AssetPlugin>()
+                .add_after::<AssetPlugin, _>(asset::plugin)
+                // Add `WindowPlugin` via `window::plugin`.
+                .disable::<WindowPlugin>()
+                .add_after::<WindowPlugin, _>(window::plugin),
```
MiniaczQ, line 25: flatten those two plugins to here

benfrankel: Okay. It's another thing we may want to re-add in the future, but that's fine.

MiniaczQ (approved):

PR #93, opened 2024-07-09
Make splash screen image explicitly linear

benfrankel:
If a user sets `DefaultPlugins.set(ImagePlugin::default_nearest())`, e.g. for a pixel art game, the splash screen image will look awful.

Explicitly specify a linear sampler for the splash screen image to guarantee that won't happen.

### DIFF
--- a/src/screen/splash/mod.rs
+++ b/src/screen/splash/mod.rs
@@ -49,6 +49,8 @@ fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
                     image: UiImage::new(asset_server.load_with_settings(
                         "embedded://bevy_template/screen/splash/splash.png",
                         |settings: &mut ImageLoaderSettings| {
+                            // Make an exception for the splash image in case
+                            // `ImagePlugin::default_nearest()` is used for pixel art.
                             settings.sampler = ImageSampler::linear();
                         },
                     )),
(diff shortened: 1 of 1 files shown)


src/screen/splash/mod.rs
```diff
@@ -42,9 +46,12 @@ fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
                         width: Val::Percent(70.0),
                         ..default()
                     },
-                    image: UiImage::new(
-                        asset_server.load("embedded://bevy_template/screen/splash/splash.png"),
-                    ),
+                    image: UiImage::new(asset_server.load_with_settings(
+                        "embedded://bevy_template/screen/splash/splash.png",
+                        |settings: &mut ImageLoaderSettings| {
+                            settings.sampler = ImageSampler::linear();
```
janhohenheim, line 52:
```suggestion
                            // This is already the default, but we are setting it explicitly
                            // because you may want to create a pixel-art game using
                            // `ImagePlugin::default_nearest()`, which would break this image
                            settings.sampler = ImageSampler::linear();
```

janhohenheim (changes requested):
I think that's a good place to educate people :)

MiniaczQ (approved):

janhohenheim (approved):

PR #95, opened 2024-07-09
Use trunk for local Wasm dev

### LINKED ISSUE #25: Tell users how they can locally run their game on Wasm

### LINKED ISSUE #49: Provide tool to run release workflow locally
This would allow users to test their changes e.g. on itch.io without having to go through the whole commit + trigger and wait for CD process, improving iteration times. This is especially useful if you only want to do the release for one particular platform, or if you pay for CD.

My repo has a bash script for this, which is better than nothing but not portable.

janhohenheim:
Resolves #49
Resolves #25

Fixes Wasm builds on Windows in general due to https://github.com/bevyengine/bevy/issues/14246


Trunk.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+# Point to our subdirectory
+target = "web/index.html"
+# This is needed in order to host the game on itch.io
+public_url = "./"
```
MiniaczQ, line 5:
```suggestion
public_url = "./"

[server]
# Required for asset loading
no_spa = true
```

web/index.html
```diff
@@ -19,14 +22,6 @@
     </div>
 
     <script type="module">
-        import './restart-audio-context.js'
```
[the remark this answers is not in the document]
MiniaczQ: Added by trunk

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 .cargo/config.toml
-/wasm
+wasm/
+dist/
```
MiniaczQ, line 4: No `wasm` generated for me and `dist` is in `web`

Trunk.toml
```diff
@@ -0,0 +1,9 @@
...
+
+[server]
+# Required for asset loading
+no_spa = true
```
MiniaczQ, line 9: can add `open = true` as well

MiniaczQ (approved):

.github/workflows/release.yaml
```diff
@@ -72,22 +72,11 @@ jobs:
 
       - name: Install dependencies
         run: |
-          cargo install wasm-bindgen-cli
+          cargo install trunk
 
-      - name: Build binary
+      - name: Build Wasm
```
benfrankel, line 77: Why rename this? Wasm is still a binary, no?

web/index.html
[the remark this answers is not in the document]
janhohenheim: T'was not

src/screen/splash.rs
```diff
@@ -42,9 +41,7 @@ fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
                         width: Val::Percent(70.0),
                         ..default()
                     },
-                    image: UiImage::new(
-                        asset_server.load("embedded://bevy_template/screen/splash/splash.png"),
-                    ),
+                    image: UiImage::new(asset_server.load("splash.png")),
```
benfrankel, line 44: There's no way to reconcile this being an embedded asset with trunk?

.gitignore
benfrankel, line 4: This will ignore any folder named `wasm/` or `dist/`. For example `src/foo/bar/dist/xyz.rs`. I don't think that's desirable.

benfrankel (changes requested):

.gitignore
janhohenheim: Fixed in the meantime

.github/workflows/release.yaml
janhohenheim: Yeah, but trunk does more than just building the binary. It also runs `wasm-bindgen`, moves assets, builds the actual `index.html`, etc. I see it more as a "Build entire Wasm package"

Trunk.toml
janhohenheim: [`afddc6b` (#95)](https://github.com/TheBevyFlock/bevy-template/pull/95/commits/afddc6ba606cb6d7c7b18d8f9e2e168b139ae5ce)

.gitignore
janhohenheim: Done

.github/workflows/release.yaml
benfrankel: I would name this `Prepare package` then.

src/screen/splash.rs
janhohenheim: [`e5f642c` (#95)](https://github.com/TheBevyFlock/bevy-template/pull/95/commits/e5f642cd0727910094b6c4559c963ea604546d6c)

benfrankel (approved):

janhohenheim:
@NiklasEi can confirm that the behavior is different when running `--no-spa` manually. Sigh. Good catch, I'll add it to the readme.

Issue #69, opened 2024-07-08
Add SFX


MiniaczQ:
Let's go for in-game music and stepping sound when the character is moving

PR #96, opened 2024-07-09
Audio

MiniaczQ:
Fixes: #69


Cargo.toml
```diff
@@ -11,6 +11,7 @@ log = { version = "0.4", features = [
     "max_level_debug",
     "release_max_level_warn",
 ] }
+rand = "0.8.5"
```
benfrankel, line 14:
```suggestion
rand = "0.8"
```

src/game/audio.rs
benfrankel, line 1: Would prefer to split this up by feature, like music vs UI sfx vs step sfx etc.

Can be done in a separate PR if that's controversial.

src/screen/playing.rs
```diff
@@ -18,15 +18,12 @@ pub(super) fn plugin(app: &mut App) {
...
 }
 
-/// Reset the camera to the default position.
-/// Not needed in this case, as we don't actually move the camera, but useful
-/// for most games.
-fn exit_playing(mut cameras: Query<&mut Transform, With<Camera>>) {
-    for mut transform in &mut cameras {
-        *transform = default();
-    }
+fn exit_playing(mut commands: Commands) {
```
benfrankel, line 24: I don't think camera transform reset should be removed. Regardless it shouldn't be removed as part of this PR.

src/ui_tools/interaction.rs
```diff
@@ -17,14 +19,21 @@ pub struct InteractionPalette {
...
+    mut commands: Commands,
 ) {
     for (interaction, (palette, mut background)) in &mut palette_query {
         *background = match interaction {
             Interaction::None => palette.none,
-            Interaction::Hovered => palette.hovered,
-            Interaction::Pressed => palette.pressed,
+            Interaction::Hovered => {
+                commands.trigger(Sfx::ButtonHover);
+                palette.hovered
```
benfrankel, line 31: These should not be coupled. Trigger the SFX in a separate system without `InteractionPalette` / `BackgroundColor` in the query, and filter for `With<Button>` or a custom marker component.

System names can be `apply_interaction_palette` and `trigger_interaction_sfx`, for example.

benfrankel (changes requested):
Some preliminary comments. May do another review pass.

src/game/audio.rs
janhohenheim: I'd leave it as-is. It's pretty clear to a user how they could split this up by themselves given the context of the rest of the template. But yeah, this is another one of these "perceived complexity vs neat packaging" tradeoffs.

src/screen/playing.rs
MiniaczQ: As mentioned in the comments, the camera is stationary in the example.
This is unnecessary

MiniaczQ: If anything, I'd be more towards making a dedicated, permanent UI camera, and a gameplay camera that gets created at the beginning

benfrankel: In the majority of games, the camera won't be stationary. And if it is, resetting its transform doesn't hurt. Multiple cameras seems unnecessary since the same camera can be used as the main game camera as well as the default UI camera.

Anyways this is all unrelated to audio.

src/game/audio.rs
benfrankel: I disagree. This is analogous to having a `game/image.rs` module that handles the duck image for the player entity as well as the splash screen image. Anyways, we can discuss this more on discord since I'm not blocking the PR on this.

MiniaczQ: I'm with Jan on this, the logic is very simple.

If we split it to two files then we have 2 separate audio modules, so we group them into a folder.
That's as much layout complexity as spawn, while the feature is way smaller.

benfrankel: No, I'm saying to split this into the existing modules by the feature they correspond to, and a new `music.rs` module because it's a new feature.

src/game/audio.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Playing a new soundtrack will overwrite the previous one.
+/// Soundtracks will loop.
+#[derive(Event)]
+pub enum Soundtrack {
+    Credits,
+    Gameplay,
+    Disable,
+}
+
+/// Play an a single sound effect.
```
benfrankel:
```suggestion
/// Play a single sound effect.
```

src/game/audio.rs
```diff
@@ -0,0 +1,77 @@
...
+        ..default()
+    };
+    commands.spawn(AudioSourceBundle { source, settings });
+}
+
+/// We mark our soundtrack entity so we can find it later.
+#[derive(Component)]
+struct SoundtrackMarker;
+
+/// Play or disable soundtrack.
```
benfrankel:
```suggestion
/// Play or disable the soundtrack.
```

src/game/audio.rs
MiniaczQ: I had to split it because of UI sfx playing interaction

src/game/audio.rs
```diff
@@ -0,0 +1,77 @@
...
+    let settings = PlaybackSettings {
+        mode: PlaybackMode::Despawn,
+        ..default()
+    };
+    commands.spawn(AudioSourceBundle { source, settings });
+}
+
+/// We mark our soundtrack entity so we can find it later.
+#[derive(Component)]
+struct SoundtrackMarker;
```
benfrankel: Consider `IsSoundtrack`. Either way fine though.

src/game/movement.rs
```diff
@@ -3,8 +3,12 @@
 //! If you want to move the player in a smoother way,
 //! consider using a [fixed timestep](https://github.com/bevyengine/bevy/pull/14223).
 
+use std::time::Duration;
+
 use bevy::prelude::*;
 
+use crate::game::audio::Sfx;
+
```
benfrankel, line 11:
```suggestion
```

benfrankel (changes requested):
Nits.

src/ui_tools/interaction.rs
```diff
@@ -17,7 +17,7 @@ pub struct InteractionPalette {
     pub pressed: Color,
 }
 
-fn apply_interaction_palette(
+fn apply_interaction(
```
benfrankel, line 20:
```suggestion
fn apply_interaction_palette(
```

src/ui_tools/interaction.rs
```diff
@@ -2,7 +2,7 @@ use bevy::prelude::*;
 
 pub(super) fn plugin(app: &mut App) {
     app.register_type::<InteractionPalette>();
-    app.add_systems(Update, apply_interaction_palette);
+    app.add_systems(Update, apply_interaction);
```
benfrankel, line 5:
```suggestion
    app.add_systems(Update, apply_interaction_palette);
```

benfrankel (changes requested):

benfrankel (approved):

janhohenheim (approved):

PR #97, opened 2024-07-09
Fix Trunk builds

janhohenheim:
Trunk only reads the `release` profile, see https://github.com/trunk-rs/trunk/issues/780

Also: trunk can already to `wasm-opt`, so no need to have a CI step for that.
I also switched from `cargo install` to `cargo binstall` to not spend 5 minutes compiling trunk

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,9 @@ opt-level = 1
 [profile.dev.package."*"]
 opt-level = 3
 
-# Enable more optimization in release mode at the cost of compile time.
+# The default profile is optimized for Wasm builds because
+# that's what [Trunk reads](https://github.com/trunk-rs/trunk/issues/605).
+# Optimize for size in wasm-release mode to reduce load times and bandwidth usage on web.
 [profile.release]
 # Compile the entire crate as one unit.
 # Significantly slows compile times, marginal improvements.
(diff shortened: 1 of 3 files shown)


benfrankel (approved):

Cargo.toml
```diff
@@ -50,21 +50,27 @@ opt-level = 1
...
-[profile.wasm-release]
-# Use release profile as default values.
-inherits = "release"
 # Optimize with size in mind (also try "s", sometimes it is better).
 # This doesn't increase compilation times compared to -O3, great improvements.
 opt-level = "z"
 # Strip all debugging information from the binary to reduce file size.
 strip = "debuginfo"
+
+# Override some settings for native builds.
```
benfrankel, line 69:
```suggestion
# Enable more optimization in release mode at the cost of compile time.
```

Cargo.toml
benfrankel, line 1: No need to move the profiles around. `release` profile can inherit from `release-native` profile, so only a simple rename is needed.

benfrankel (changes requested):

Cargo.toml
benfrankel: Nevermind, this is wrong! `release` profile contains the default values we want in `release-native`.

Cargo.toml
benfrankel: Will make more sense to restore this comment when trunk is fixed upstream and this whole workaround is reverted.

PR #100, opened 2024-07-09
Add splash screen fade in/out animation

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy-template/issues/91.

There's no fade in for the title screen, so it kinda harshly snaps into existence.

### DIFF
--- a/src/screen/splash.rs
+++ b/src/screen/splash.rs
@@ -10,8 +10,8 @@ use crate::ui_tools::prelude::*;
 
 pub(super) fn plugin(app: &mut App) {
     // Spawn splash screen.
-    app.add_systems(OnEnter(Screen::Splash), spawn_splash);
     app.insert_resource(ClearColor(SPLASH_BACKGROUND_COLOR));
+    app.add_systems(OnEnter(Screen::Splash), spawn_splash);
 
     // Animate splash screen.
     app.add_systems(
--- a/src/screen/splash.rs
+++ b/src/screen/splash.rs
@@ -93,9 +93,7 @@ impl UiImageFadeInOut {
         let fade = self.fade_duration / self.total_duration;
 
         // Regular trapezoid-shaped graph, flat at the top with alpha = 1.0.
-        let t = ((1.0 - (2.0 * t - 1.0).abs()) / fade).min(1.0);
-
-        t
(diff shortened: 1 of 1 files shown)


src/screen/splash.rs
```diff
@@ -55,11 +66,48 @@ fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
...
+        let fade = self.fade_duration / self.total_duration;
+
+        // Regular trapezoid-shaped graph, flat at the top with alpha = 1.0.
+        ((1.0 - (2.0 * t - 1.0).abs()) / fade).min(1.0)
+    }
+}
+
+fn tick_fade_in_out(time: Res<Time>, mut animation_query: Query<&mut UiImageFadeInOut>) {
+    for mut anim in &mut animation_query {
+        anim.t += time.delta_seconds();
```
MiniaczQ, line 102: When working with curves you usually talk about `t` in `0..=1` range, I think the scaling should happen here

benfrankel: Yeah I had it like that at one point. Scaling later makes it more ergonomic to add like an `anim.rate` field that you can adjust while the animation plays. Ofc we're not doing that, but I had to choose between scale early or scale late, so that was why I chose scale late.

MiniaczQ: Removed my comment, didn't see you replied until I refreshed the page 😛

MiniaczQ (changes requested):
Waiting for math dev on reply about curves

src/screen/splash.rs
```diff
@@ -9,10 +9,19 @@ use super::Screen;
 use crate::{ui_tools::prelude::*, AppStep};
 
 pub(super) fn plugin(app: &mut App) {
-    app.add_systems(OnEnter(Screen::Splash), spawn_splash);
+    // Spawn splash screen.
     app.insert_resource(ClearColor(SPLASH_BACKGROUND_COLOR));
```
MiniaczQ, line 13:
```suggestion
    app.insert_resource(ClearColor(SPLASH_BACKGROUND_COLOR));
    // Spawn splash screen.
```

MiniaczQ (approved):

Issue #94, opened 2024-07-09
Show how to use animations


janhohenheim:
The ducky has some cute animations: https://caz-creates-games.itch.io/ducky-2

PR #110, opened 2024-07-11
Duck animation

MiniaczQ:
Fixes: #94


src/lib.rs
```diff
@@ -69,8 +66,6 @@ impl Plugin for AppPlugin {
 /// call above.
 #[derive(SystemSet, Debug, Clone, Copy, Eq, PartialEq, Hash)]
 enum AppSet {
-    /// Tick timers.
-    TickTimers,
```
benfrankel, line 73: Why remove this?

src/game/movement.rs
```diff
@@ -120,36 +119,122 @@ fn update_facing(mut player_query: Query<(&MovementController, &mut Sprite)>) {
...
+            return;
+        }
+        match self.state {
+            PlayerAnimationState::Idle => {
+                self.frame = (self.frame + 1) % Self::IDLE_FRAMES;
+            }
+            PlayerAnimationState::Walking => {
+                self.frame = (self.frame + 1) % Self::WALKING_FRAMES;
+            }
+        }
```
benfrankel, line 215:
```suggestion
        self.frame = (self.frame + 1) % match self.state {
            PlayerAnimationState::Idle => Self::IDLE_FRAMES,
            PlayerAnimationState::Walking => Self::WALKING_FRAMES,
        };
```

src/game/movement.rs
```diff
@@ -120,36 +119,122 @@ fn update_facing(mut player_query: Query<(&MovementController, &mut Sprite)>) {
...
+            state: PlayerAnimationState::Walking,
+        }
+    }
+
+    pub fn new() -> Self {
+        Self::idle()
+    }
+
+    /// Update animation timers.
+    pub fn update_timer(&mut self, delta: Duration) {
```
benfrankel, line 203: Would name this `PlayerAnimation::tick` to match `Timer::tick`.

src/game/movement.rs
```diff
@@ -30,14 +30,13 @@ pub(super) fn plugin(app: &mut App) {
     // Update facing based on controls.
     app.add_systems(Update, update_facing.in_set(AppSet::Update));
 
-    // Trigger step sound effects based on controls.
-    app.register_type::<StepSfx>();
+    // Animate and play sound effects on controls.
```
benfrankel, line 33:
```suggestion
    // Animate and play sound effects based on controls.
```

benfrankel (changes requested):

src/game/movement.rs
MiniaczQ: This matches `update_state`, which is better since they're similar methods on the same object

src/lib.rs
MiniaczQ: splash timer runs before everything else, but animation timer runs after inputs
so this system set would only be used by a single system, i settled for chain

src/game/movement.rs
benfrankel: I'd argue that update animation should be split into two systems, one to tick the timer and one to update the animation, but this is fine I guess.

src/lib.rs
benfrankel: I'd argue that update animation should be split into two systems, one to tick the timer and one to update the animation, but this is fine I guess.

src/game/movement.rs
benfrankel: They do very different things so I'm not personally convinced, but alright.

src/lib.rs
janhohenheim: I agree with @benfrankel here\

src/game/movement.rs
```diff
@@ -120,36 +119,118 @@ fn update_facing(mut player_query: Query<(&MovementController, &mut Sprite)>) {
...
+#[derive(Reflect, PartialEq)]
+pub enum PlayerAnimationState {
+    Idle,
+    Walking,
 }
 
-fn tick_step_sfx(time: Res<Time>, mut step_query: Query<&mut StepSfx>) {
-    for mut step in &mut step_query {
-        step.cooldown_timer.tick(time.delta());
+impl PlayerAnimation {
```
janhohenheim, line 171: Could we move the animation stuff to a new file?

janhohenheim (changes requested):

src/lib.rs
MiniaczQ: Done

src/game/movement.rs
MiniaczQ: Done

src/game/movement.rs
MiniaczQ: Well I'm not convinced we should reflect timer API in animation controller, because it contains one :p

janhohenheim:
Maybe we could spend a few words explaining what a texture atlas is? I know it's a ubiquitous term, but a beginner might still not have heard of it yet.
Not blocking on that though.

src/game/mod.rs
```diff
@@ -2,11 +2,17 @@
...
 pub mod audio;
 mod movement;
 pub mod spawn;
 
 pub(super) fn plugin(app: &mut App) {
-    app.add_plugins((audio::plugin, movement::plugin, spawn::plugin));
+    app.add_plugins((
+        audio::plugin,
+        movement::plugin,
+        animation::plugin,
```
benfrankel, line 15:
```suggestion
        animation::plugin,
        audio::plugin,
        movement::plugin,
```
Alphabetizing barring a reason not to :P

src/game/spawn/player.rs
```diff
@@ -24,19 +23,36 @@ pub struct SpawnPlayer;
...
+    let texture_atlas_layout = texture_atlas_layouts.add(layout);
+    let player_animation = PlayerAnimation::new();
+
     commands.spawn((
         Name::new("Player"),
         Player,
         SpriteBundle {
             texture: images[&ImageAsset::Ducky].clone_weak(),
-            transform: Transform::from_scale(Vec3::splat(0.5)),
+            transform: Transform::from_scale(Vec3::splat(8.0)),
```
benfrankel, line 45: It was already like this before, but technically we don't want to scale Z. But I don't think it actually does anything in 2D so w/e.

src/screen/splash.rs
```diff
@@ -27,10 +29,8 @@ pub(super) fn plugin(app: &mut App) {
     app.add_systems(OnExit(Screen::Splash), remove_splash_timer);
     app.add_systems(
         Update,
-        (
-            tick_splash_timer.in_set(AppSet::TickTimers),
-            check_splash_timer.in_set(AppSet::Update),
-        )
+        (tick_splash_timer, check_splash_timer)
+            .chain()
```
benfrankel, line 33:
```suggestion
        (
            tick_splash_timer.in_set(AppSet::TickTimers),
            check_splash_timer.in_set(AppSet::Update),
        )
```

benfrankel (changes requested):
Few nits.

src/screen/splash.rs
MiniaczQ: I thought I fixed this 🤔

src/game/animation.rs
```diff
@@ -0,0 +1,161 @@
...
+        if animation.changed() {
+            atlas.index = animation.get_atlas_index();
+        }
+    }
+}
+
+/// If the player is moving, play a step sound effect synchronized with the animation.
+fn trigger_step_sfx(mut commands: Commands, mut step_query: Query<&PlayerAnimation>) {
+    for animation in &mut step_query {
+        if animation.state == PlayerAnimationState::Walking
```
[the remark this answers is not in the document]
MiniaczQ: I tried removing `PartialEq` if possible, but we rely on it later in code :/
I'll just keep it as is then

src/game/spawn/player.rs
MiniaczQ: done

src/game/spawn/player.rs
```diff
@@ -24,19 +23,36 @@ pub struct SpawnPlayer;
...
+    let texture_atlas_layout = texture_atlas_layouts.add(layout);
+    let player_animation = PlayerAnimation::new();
+
     commands.spawn((
         Name::new("Player"),
         Player,
         SpriteBundle {
             texture: images[&ImageAsset::Ducky].clone_weak(),
-            transform: Transform::from_scale(Vec3::splat(0.5)),
+            transform: Transform::from_scale(Vec2::splat(8.0).extend(0.0)),
```
benfrankel, line 45:
```suggestion
            transform: Transform::from_scale(Vec2::splat(8.0).extend(1.0)),
```

benfrankel (approved):

Issue #71, opened 2024-07-08
Add asset loading stage


janhohenheim:

MiniaczQ:
Out of scope for 0.14 template

janhohenheim:
Re-adding this issue. Quoting @alice-i-cecile:
> BTW, for audio (especially on web), you really need to load the various sounds into memory ahead of time or you'll get nasty stuttering

PR #111, opened 2024-07-11
Preload game assets

MiniaczQ:
Fixes: #71

### DIFF
--- a/src/game/assets.rs
+++ b/src/game/assets.rs
@@ -1,69 +1,127 @@
-use bevy::prelude::*;
+use std::ops::Deref;
+
+use bevy::{prelude::*, utils::HashMap};
+
+#[derive(PartialEq, Eq, Hash, Reflect)]
+pub enum ImageAsset {
+    Ducky,
+}
 
 #[derive(Resource, Reflect)]
 pub struct ImageAssets {
-    pub ducky: Handle<Image>,
+    pub assets: HashMap<ImageAsset, Handle<Image>>,
 }
 
 impl ImageAssets {
     pub fn new(asset_server: &AssetServer) -> Self {
-        Self {
-            ducky: asset_server.load("images/ducky.png"),
-        }
+        let mut assets = HashMap::new();
+
+        assets.insert(ImageAsset::Ducky, asset_server.load("images/ducky.png"));
+
+        Self { assets }
     }
 
     pub fn all_loaded(&self, assets: &Assets<Image>) -> bool {
-        assets.contains(&self.ducky)
+        self.assets
+            .iter()
+            .all(|(_, handle)| assets.contains(handle))
     }
+}
+
+impl Deref for ImageAssets {
+    type Target = HashMap<ImageAsset, Handle<Image>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.assets
+    }
+}
+
+#[derive(PartialEq, Eq, Hash, Reflect)]
+pub enum SfxAsset {
+    ButtonHover,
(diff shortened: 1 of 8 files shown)


src/game/audio/soundtrack.rs
```diff
@@ -1,23 +1,24 @@
...
         commands.entity(entity).despawn();
     }
-    let path = match event {
+    let source = match event {
         Soundtrack::Disable => {
             return;
         }
-        Soundtrack::Credits => "audio/soundtracks/Monkeys Spinning Monkeys.ogg",
-        Soundtrack::Gameplay => "audio/soundtracks/Fluffing A Duck.ogg",
+        Soundtrack::Credits => soundtracks.credits.clone(),
```
alice-i-cecile, line 19: These can be cheaper weak clones: you're already holding onto a durable handle.

src/game/audio/sfx.rs
```diff
@@ -28,13 +25,8 @@ pub enum Sfx {
...
-fn random_step() -> &'static str {
-    [
-        "audio/sfx/step1.ogg",
-        "audio/sfx/step2.ogg",
-        "audio/sfx/step3.ogg",
-        "audio/sfx/step4.ogg",
-    ]
-    .choose(&mut rand::thread_rng())
-    .unwrap()
+fn random_step(sfxs: &SfxAssets) -> &Handle<AudioSource> {
```
alice-i-cecile, line 28: This should be a methods on `SfxAssets` :)

src/game/assets.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn all_loaded(&self, assets: &Assets<Image>) -> bool {
+        assets.contains(&self.ducky)
+    }
+}
+
+#[derive(Resource, Reflect)]
+pub struct SfxAssets {
+    pub button_hover: Handle<AudioSource>,
+    pub button_press: Handle<AudioSource>,
+    pub step1: Handle<AudioSource>,
```
alice-i-cecile, line 24: These make more sense as a HashSet of handles.

janhohenheim: You mean a `HashMap`, I suppose?

src/game/audio/sfx.rs
janhohenheim: It seems weird to me to couple in-game SFX logic to the SFX asset struct. IMO this would be a bad precedent.

janhohenheim (changes requested):
Agreed with 2/3 suggestions by Alice, LGTM otherwise :)

src/screen/loading.rs
```diff
@@ -0,0 +1,44 @@
+//! A credits screen that can be accessed from the title screen.
```
janhohenheim, line 1: I think you forgot to update this comment ;)

janhohenheim (changes requested):

src/screen/loading.rs
MiniaczQ: 😭

src/game/assets.rs
```diff
@@ -1,69 +1,127 @@
...
+    }
+}
+
+impl Deref for SfxAssets {
+    type Target = HashMap<SfxAsset, Handle<AudioSource>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.assets
     }
 }
```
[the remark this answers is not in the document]
MiniaczQ: I didn't know std had `Deref` derive, I'll do that

janhohenheim (approved):

PR #127, opened 2024-07-12
Yeet sccache

janhohenheim:


janhohenheim:
Do not merge:
image

Looks like sccache is not doing anything

janhohenheim:
Update: this now yeets sccache instead

benfrankel (approved):

mpwoz:
Just curious, since i'm in the process of trying to add this to my build

https://github.com/actions/cache/blob/main/examples.md#rust---cargo

Is it an issue with sscache, or did caching in general not result in any speedup? Could save me some time to know if it's not worth even trying to add

benfrankel:
Caching is big:

2024-07-12_1720807524_200x138

That's two consecutive successful releases, the older one with an empty cache.

benfrankel:
Sccache may be good as well, but it didn't seem to work on first attempt so we're planning on revisiting it later.

janhohenheim:
@mpwoz definitely use caching!
We didn't get sccache to work, but are happily using [cargo-cache](https://github.com/Leafwing-Studios/cargo-cache) in the meantime :)

PR #130, opened 2024-07-12
Improve CI some more

benfrankel:


.github/workflows/ci.yaml
```diff
@@ -11,86 +11,104 @@ env:
 
 jobs:
   # Run unit tests.
-  test:
-    name: Test Suite
+  unit-test:
+    name: Unit Tests
```
janhohenheim, line 15: Misnomer: also includes integration tests, regression tests, smoke tests, benchmarks, etc.
This comment is for all instances of this rename

janhohenheim (changes requested):

.github/workflows/ci.yaml
benfrankel: Alright. What should I call this to distinguish it from doc tests?

janhohenheim: But it also runs doc tests! The second one is more of a `Check Docs` or `Validate Docs`

benfrankel: Okay. The manual page was confusing:
```
       --all-targets
           Test all targets. This is equivalent to specifying --lib --bins --tests --benches --examples.

       --doc
           Test only the library’s documentation. This cannot be mixed with other target options.
```

janhohenheim: Wait, hold up. You're right, it does not run doc tests

janhohenheim: `Tests` vs `Doc Tests` is good for me then.

benfrankel: https://github.com/rust-lang/cargo/issues/6669 :smiling_face_with_tear:

.github/workflows/ci.yaml
```diff
@@ -10,87 +10,87 @@ env:
...
-      - name: Run cargo test
-        run: cargo test --all-targets
+      - name: Populate target directory from cache
+        uses: Leafwing-Studios/cargo-cache@v1
+
+      - name: Run tests
+        run: |
+          cargo test --workspace --all-targets --all-features
+          # TODO: Workaround for https://github.com/rust-lang/cargo/issues/6669
+          cargo test --workspace --docs --all-features
```
janhohenheim, line 35:
```suggestion
          # Workaround for https://github.com/rust-lang/cargo/issues/6669
          cargo test --workspace --docs
```
This reads as if the user needs to do something

janhohenheim (changes requested):

.github/workflows/ci.yaml
benfrankel: Yeah fair enough. This is how I mark it in my own projects so I know to remove the workaround when it's no longer necessary.

.github/workflows/ci.yaml
```diff
@@ -10,87 +10,87 @@ env:
...
+      - name: Install dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
 
-      - name: Run cargo doc
-        run: cargo doc --no-deps
+      - name: Populate target directory from cache
+        uses: Leafwing-Studios/cargo-cache@v1
+
+      - name: Check documentation
+        run: cargo doc --workspace --all-features
```
janhohenheim, line 96:
```suggestion
        run: cargo doc --workspace --all-features --document-private-items
```
This is what we are interested in, after all

janhohenheim (changes requested):

.github/workflows/ci.yaml
```diff
@@ -10,87 +10,87 @@ env:
...
-      - name: Run cargo test
-        run: cargo test --all-targets
+      - name: Populate target directory from cache
+        uses: Leafwing-Studios/cargo-cache@v1
+
+      - name: Run tests
+        run: |
+          cargo test --workspace --all-targets --all-features
+          # Workaround for https://github.com/rust-lang/cargo/issues/6669
+          cargo test --workspace --docs
```
janhohenheim, line 35:
```suggestion
          cargo test --workspace --doc
```

janhohenheim (changes requested):

.github/workflows/ci.yaml
benfrankel: Oops.

janhohenheim (approved):

.github/workflows/ci.yaml
```diff
@@ -10,87 +10,87 @@ env:
...
+      - name: Install dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev
 
-      - name: Run cargo doc
-        run: cargo doc --no-deps
+      - name: Populate target directory from cache
+        uses: Leafwing-Studios/cargo-cache@v1
+
+      - name: Check documentation
+        run: cargo doc --workspace --all-features --document-private-items
```
janhohenheim, line 96:
```suggestion
        run: cargo doc --workspace --all-features --document-private-items --no-deps
```

janhohenheim (changes requested):

janhohenheim (changes requested):
Add `RUSTDOCFLAGS='-D warnings'` to the docs step.
You're about to whether there is a `--deny-warnings` or similar.
No: https://github.com/rust-lang/cargo/issues/8424

janhohenheim (approved):

PR #132, opened 2024-07-13
Use matrix in release.yaml

benfrankel:
May make https://github.com/TheBevyFlock/bevy-template/issues/40 easier, if their workflows are similar enough.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,7 +16,6 @@ on:
 
 # Configure constants for this workflow.
 env:
-  # This may become unnecessary: https://github.com/rust-lang/cargo/issues/6790
   # The base filename of the binary produced by `cargo build`.
   BINARY: bevy_template
   # The name to use for the packaged application produced by this workflow.
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,7 +35,7 @@ env:
   USE_GIT_LFS: false
 
 jobs:
-  # Extract the version number from the pushed tag.
+  # Determine the version number for this workflow.
   get-version:
     runs-on: ubuntu-latest
     steps:
@@ -134,10 +134,13 @@ jobs:
           - platform: macos
             targets: x86_64-apple-darwin aarch64-apple-darwin
             profile: release-native
-            package_ext: .dmg
             out_dir_suffix: .app/Contents/MacOS
+            package_ext: .dmg
             runner: macos-latest
     runs-on: ${{ matrix.runner || 'ubuntu-latest' }}
+    defaults:
+      run:
+        shell: bash
     steps:
       - name: Set up environment
         run: |
@@ -220,7 +223,8 @@ jobs:
             </plist>
           EOF
 
-      - name: Finish package
+      - name: Finish package (non-Windows)
+        if: ${{ matrix.platform != 'windows' }}
         working-directory: tmp/package
         run: |
           if [ '${{ matrix.platform }}' == 'macos' ]; then
@@ -229,6 +233,12 @@ jobs:
           else
             zip --recurse-paths '${{ env.PACKAGE }}${{ matrix.package_ext }}' '${{ env.PACKAGE_NAME }}'
           fi
+
+      - name: Finish package (Windows)
+        if: ${{ matrix.platform == 'windows' }}
+        working-directory: tmp/package
+        shell: pwsh
+        run: Compress-Archive -Path '${{ env.PACKAGE_NAME }}' -DestinationPath '${{ env.PACKAGE }}${{ matrix.package_ext }}'
 
       - name: Upload package to artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -155,12 +155,12 @@ jobs:
         with:
           targets: ${{ matrix.target }}
 
+      - name: Populate target directory from cache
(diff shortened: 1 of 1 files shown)


benfrankel:
Latest commit actually makes this not fix https://github.com/TheBevyFlock/bevy-template/issues/117. May revert, we'll see.

benfrankel:
> My current take: Keep the simpler approach for now. `macos` still finishes 3min faster than `windows`, and only like 30sec slower than `linux` in the few release runs that I've tried.
> If `macos` becomes the bottleneck in the future, it can be re-evaluated.

2024-07-12_1720840269_980x334

.github/workflows/release.yaml
```diff
@@ -25,178 +26,130 @@ env:
...
   get-version:
     runs-on: ubuntu-latest
     steps:
       - name: Get version number from tag
         id: tag
-        run: |
-          echo "tag=${GITHUB_REF#refs/tags/}" >> "${GITHUB_OUTPUT}"
+        run: echo "tag=${GITHUB_REF#refs/tags/}" >> "${GITHUB_OUTPUT}"
     outputs:
       version: ${{ inputs.version || steps.tag.outputs.tag }}
```
BD103, line 46:
```suggestion
      # Use inputted version if this is a `workflow_dispatch`, else use the current tag.
      version: ${{ inputs.version || steps.tag.outputs.tag }}
```

The `||` syntax is a bit strange (and unlike other languages), so a comment might help explain that it functions similar to a ternary operator.

.github/workflows/release.yaml
```diff
@@ -25,178 +26,130 @@ env:
...
-      - name: Set PACKAGE environment variable
-        run: |
-          echo "PACKAGE=${{ env.PACKAGE_NAME }}-${{ env.VERSION }}-${{ env.PLATFORM }}" >> "${GITHUB_ENV}"
-
-      - name: Set OUT_DIR environment variable
+      - name: Set up environment
         run: |
-          echo "OUT_DIR=build/${{ env.PACKAGE }}/${{ env.PACKAGE_NAME }}" >> "${GITHUB_ENV}"
+          echo 'PACKAGE=${{ env.PACKAGE_NAME }}-${{ env.PLATFORM }}' >> "${GITHUB_ENV}"
+          echo 'OUT_DIR=tmp/package/${{ env.PACKAGE_NAME }}' >> "${GITHUB_ENV}"
```
BD103, line 61: This mixes both single quotes and double quotes. I would prefer the file prefer one and drop the other, except for special occasions.

- No quotes: similar to double quotes, but a chance a string will be evaluated as Bash.
- Single quotes: forces the contained expression to be a string, but does not support Bash `$VARIABLE` evaluation. (But `${{ expr }}` still works, because that's Github Actions.)
- Double quotes: forces the contained expression to be a string, and does support Bash `$VARIABLE` evaluation.

.github/workflows/release.yaml
```diff
@@ -25,178 +26,130 @@ env:
...
       VERSION: ${{ needs.get-version.outputs.version }}
-
+    strategy:
+      matrix:
+        include:
+          - platform: linux
+            targets: x86_64-unknown-linux-gnu
+            profile: release-native
+            features: bevy/wayland
+            package_ext: .zip
```
BD103, line 125: I'm not sure I'm a fan of `ubuntu-latest` being the default, especially since WASM is a separate job and is the only other target that could use Linux.

```suggestion
          - platform: linux
            targets: x86_64-unknown-linux-gnu
            profile: release-native
            features: bevy/wayland
            package_ext: .zip
            runner: ubuntu-latest
```

.github/workflows/release.yaml
benfrankel: This is intentional. The expressions like `${{ expr }}` get evaluated before the command is passed to the shell, so if the evaluated expression contains backslashes or `$` or other special characters, it would cause issues within double quotes. Single quotes make more sense.

.github/workflows/release.yaml
benfrankel: I'm considering combining the web job with the matrix now that it's possible (since there's no longer an intermediate step to upload a binary). Android would also use `ubuntu-latest`, and I think iOS would use `macos-latest`.

Not opposed to making it explicit, though.

.github/workflows/release.yaml
BD103: Ok, sounds good!

.github/workflows/release.yaml
benfrankel: Added comment.

.github/workflows/release.yaml
```diff
@@ -1,21 +1,22 @@
...
       version:
-        description: 'Version - in the form of v1.2.3'
+        description: 'Version number in the format `v1.2.3`'
         required: true
         type: string
 
 # Configure constants for this workflow.
 env:
-  # TODO: This may become unnecessary: https://github.com/rust-lang/cargo/issues/6790
+  # This may become unnecessary: https://github.com/rust-lang/cargo/issues/6790
```
janhohenheim, line 19:
```suggestion
```
I think this comment is more confusing than helpful to a user. Better to setup an issue imo.

janhohenheim (changes requested):

.github/workflows/release.yaml
benfrankel: Okay, fair enough.

janhohenheim (approved):

PR #133, opened 2024-07-13
Make itch.io username optional and drop itch related actions if not set

Sapein:
This will make the itch.io username optional when using `cargo generate`. If it's not supplied (left blank) then `release.yaml` will have anything related to itch.io excluded when it's created.


.github/workflows/release.yaml
```diff
@@ -22,7 +22,9 @@ env:
   PACKAGE_NAME: {{project-name}}
   # The itch.io page to upload to, in the format: `user-name/project-name`.
   # Comment this out to disable.
+  {%- if itch_username != "" -%}
```
janhohenheim, line 25: Can we instead comment it out? That way, a user can still activate itch support later

.github/workflows/release.yaml
```diff
@@ -351,6 +353,9 @@ jobs:
           tag: ${{ env.VERSION }}
           overwrite: true
 
+{%- endraw -%}
+{%- if itch_username != "" -%}
```
janhohenheim, line 357: I'd remove this `if`. The workflow already handles a commented out `ITCH_TARGET` correctly. If we leave it in, a user can still add itch support later.

cargo-generate.toml
```diff
@@ -2,5 +2,6 @@
 include = ["Cargo.toml", "Cargo.lock", ".github/workflows/*", "*.rs"]
 
 [placeholders.itch_username]
-prompt = "Enter your itch.io username"
+prompt = "Enter your itch.io username. (Leave Blank to disable itch.io upload support)"
```
janhohenheim, line 5:
```suggestion
prompt = "Enter your itch.io username. Leave blank to disable itch.io uploads."
```

.github/workflows/release.yaml
```diff
@@ -22,7 +22,9 @@ env:
   PACKAGE_NAME: {{project-name}}
   # The itch.io page to upload to, in the format: `user-name/project-name`.
   # Comment this out to disable.
+  {%- if itch_username != "" -%}
   ITCH_TARGET: {{itch_username}}/{{project-name}}
+  {%- endif -%}
```
benfrankel, line 27: Would it be possible to make this line `#ITCH_TARGET:` when no username is provided, so that there's a clear place to add the itch info if the user chooses to do so later on?

cargo-generate.toml
```diff
@@ -2,5 +2,6 @@
...
 type = "string"
+default = ""
```
janhohenheim, line 7:
```suggestion
default = ""

```
Do you not have `.editorconfig` support enabled in your IDE? If you're using VSCode, I advise you to download the extension ;)

janhohenheim (changes requested):

.github/workflows/release.yaml
```diff
@@ -31,7 +33,7 @@ env:
   # https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage
   USE_GIT_LFS: false
 
-{% raw -%}
+{%- raw -%}
```
benfrankel, line 36: Curious what this `raw` block does?

.github/workflows/release.yaml
```diff
@@ -351,6 +353,9 @@ jobs:
           tag: ${{ env.VERSION }}
           overwrite: true
 
+{%- endraw -%}
+{%- if itch_username != "" -%}
+{%- raw -%}
```
benfrankel, line 358: This shouldn't be necessary. The release workflow already checks if the `ITCH_TARGET` env variable is provided, and if not, doesn't upload to itch.io.

In fact this section should be left in, to allow the user to enable itch.io later on.

benfrankel (changes requested):

.github/workflows/release.yaml
janhohenheim: It means that the variables that come later are not `liquid` variables, and thus should not be expanded by `cargo generate`

.github/workflows/release.yaml
Sapein: I can do an else statement that has the same text, but commented out if you want. That would be the best solution

.github/workflows/release.yaml
Sapein: The raw block just means: "Don't modify what's in this block". If I don't include it, then nothing will work as the syntax for actions conflicts with the templating syntax.

.github/workflows/release.yaml
benfrankel: (redundant with Jan's comment)

.github/workflows/release.yaml
benfrankel: (redundant with Jan's comment)

cargo-generate.toml
Sapein: I use RustRover

.github/workflows/release.yaml
benfrankel: Got it, thanks.

cargo-generate.toml
janhohenheim: @Sapein in that case, could you activate the [EditorConfig](https://www.jetbrains.com/help/idea/editorconfig.html) plugin? That way, we all use the same formatting style :)

.github/workflows/release.yaml
```diff
@@ -22,7 +22,11 @@ env:
   PACKAGE_NAME: {{project-name}}
   # The itch.io page to upload to, in the format: `user-name/project-name`.
   # Comment this out to disable.
+  {%- if itch_username != "" -%}
   ITCH_TARGET: {{itch_username}}/{{project-name}}
+  {%- else -%}
+  #ITCH_TARGET: {{itch_username}}/{{project-name}}
```
janhohenheim, line 28:
```suggestion
  # ITCH_TARGET: {{itch_username}}/{{project-name}}
```

.github/workflows/release.yaml
```diff
@@ -404,4 +408,4 @@ jobs:
                 builds/"${channel}"/* \
                 ${{ env.ITCH_TARGET }}:"${channel}"
           done
-{%- endraw %}
+{%- endraw -%}
```
janhohenheim, line 411:
```suggestion
{%- endraw -%}

```

cargo-generate.toml
```diff
@@ -2,5 +2,6 @@
 include = ["Cargo.toml", "Cargo.lock", ".github/workflows/*", "*.rs"]
 
 [placeholders.itch_username]
-prompt = "Enter your itch.io username"
+prompt = "Enter your itch.io username. Leave Blank to disable itch.io upload."
```
janhohenheim, line 5:
```suggestion
prompt = "Enter your itch.io username. Leave blank to disable itch.io upload."
```

janhohenheim (changes requested):

janhohenheim (approved):

PR #136, opened 2024-07-13
Split out doc tests in CI

benfrankel:


benfrankel:
This PR works now! Investigating whether the fix can be upstreamed.

benfrankel:
Closing in favor of potential upstream fix: https://github.com/Leafwing-Studios/cargo-cache/pull/17.

We can push a `cargo update` PR for now as a workaround.

PR #146, opened 2024-07-15
Rename to `bevy_quickstart` internally

benfrankel:


.github/workflows/release.yaml
```diff
@@ -17,12 +17,12 @@ on:
...
   # The base filename of the binary produced by `cargo build`.
-  BINARY: bevy_template
+  BINARY: bevy_quickstart
   # The name to use for the packaged application produced by this workflow.
-  PACKAGE_NAME: bevy-template
+  PACKAGE_NAME: bevy_quickstart
   # The itch.io page to upload to, in the format: `user-name/project-name`.
   # Comment this out to disable.
-  ITCH_TARGET: the-bevy-flock/bevy-template
+  ITCH_TARGET: the-bevy-flock/bevy-quickstart
```
janhohenheim, line 25: Is it fine if the `cargo generate` template assumes this has the same name as the binary?

benfrankel: I assume you mean the binary in the released package, aka `PACKAGE_NAME`? If so, it's probably a reasonable default, and the user can tinker with it later.

janhohenheim (approved):

PR #151, opened 2024-07-17
Clean up metadata of readmes

janhohenheim:
- Move stuff into a `docs` directory
- Change some names

I'll change the actual content of the files in a follow-up


benfrankel:
Eh I think uppercase names for README and LICENSE are more standard. The moves to `docs/` look good though.

janhohenheim:
@benfrankel more standard, yes, but also more annoying.

benfrankel:
I'd opt for standard in this case. Every repo has license and readme files, and they're always uppercase.

janhohenheim:
@benfrankel I've renamed the files.

README.md
```diff
@@ -23,9 +23,14 @@ Navigate to the top of [this GitHub repository](https://github.com/TheBevyFlock/
 ![example](TODO)
 
 Next, clone your new Github repository to a local repository and create an `Initial commit` with the following changes:
-- Delete `LICENSE` and `README` files.
+
+- Delete the following.
+  - `README.md`
+  - All `LICENSE-*.txt` files
+  - the `docs` directory
```
benfrankel, line 30:
```suggestion
- Delete `LICENSE`, `README`, and `docs/` files.
```
I think it looks less scary this way than nested list.

benfrankel (changes requested):
I think header of `docs/workflows.md` needs to be changed from `CI / CD` as well.

janhohenheim:
> I think header of `docs/workflows.md` needs to be changed from `CI / CD` as well.

I intentionally left that alone since I'm changing the contents next.

benfrankel (approved):

PR #154, opened 2024-07-17
Merge `main` into `cargo-generate`

janhohenheim:


.github/workflows/release.yaml
```diff
@@ -153,6 +153,56 @@ jobs:
 
       - name: Add app metadata to package (MacOS)
         if: ${{ matrix.platform == 'macos' }}
+        run: |
+          Compress-Archive -Path "${{ env.PACKAGE_NAME }}" -DestinationPath "../${{ env.PACKAGE }}.zip"
```
benfrankel, line 157: Merge broke here somehow, this isn't correct :fearful:

benfrankel (changes requested):

.github/workflows/release.yaml
janhohenheim: [`6f9a860` (#154)](https://github.com/TheBevyFlock/bevy_quickstart/pull/154/commits/6f9a860fcdb1e95922bcb6aed512d9572408fbd5)

benfrankel (approved):
I don't _see_ anything crazy. Approving and we can test it out.

PR #162, opened 2024-07-18
Add note on rust-analyzer

janhohenheim:


benfrankel (approved):
Looks reasonable to me.

Issue #168, opened 2024-07-18
Use LoadState instead of just checking if handle exists


paul-hansen:
Shouldn't [these lines](https://github.com/TheBevyFlock/bevy_quickstart/blob/f0bbb01a862ee8261c92ad30d7937deecca96abf/src/game/assets.rs#L71C9-L71C65) be checking the asset's [load state](https://docs.rs/bevy/latest/bevy/asset/struct.AssetServer.html#method.load_state) not just if the handles exist? I thought handles would exist immediately but not actually be loaded yet.

Something like (untested):
```rust
pub fn all_loaded(&self, assets: &AssetServer) -> bool {
    self.0.iter().all(|(_, handle)| assets.load_state(handle.into()) == LoadState::Loaded)
}
```
Or perhaps even use the new [is_loaded_with_dependencies](https://docs.rs/bevy/0.14.0/bevy/asset/struct.AssetServer.html#method.is_loaded_with_dependencies). Though it wouldn't really do anything different for the example assets.

I don't know for sure, been away from Bevy for a bit and was reading the code here to catch up. Doesn't look like the examples have been updated in the unoffical cheatbook either which is where I got my previous info: https://bevy-cheatbook.github.io/assets/ready.html

PR #169, opened 2024-07-18
Refactor some asset code

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/168.

### DIFF
--- a/src/game/assets.rs
+++ b/src/game/assets.rs
@@ -5,18 +5,14 @@ use bevy::{
 };
 
 pub(super) fn plugin(app: &mut App) {
-    app.register_type::<(
-        AssetMap<ImageKey>,
-        AssetMap<SfxKey>,
-        AssetMap<SoundtrackKey>,
-    )>();
-    app.add_systems(Startup, preload_assets);
-}
+    app.register_type::<AssetMap<ImageKey>>();
+    app.init_resource::<AssetMap<ImageKey>>();
 
-fn preload_assets(mut commands: Commands, asset_server: Res<AssetServer>) {
-    commands.insert_resource(ImageKey::load(&asset_server));
-    commands.insert_resource(SfxKey::load(&asset_server));
-    commands.insert_resource(SoundtrackKey::load(&asset_server));
+    app.register_type::<AssetMap<SfxKey>>();
+    app.init_resource::<AssetMap<SfxKey>>();
+
+    app.register_type::<AssetMap<SoundtrackKey>>();
+    app.init_resource::<AssetMap<SoundtrackKey>>();
 }
 
 #[derive(PartialEq, Eq, Hash, Reflect)]
@@ -27,16 +23,16 @@ pub enum ImageKey {
 impl AssetKey for ImageKey {
     type Asset = Image;
 
-    fn load(asset_server: &AssetServer) -> AssetMap<Self> {
-        AssetMap(HashMap::from([(
-            ImageKey::Ducky,
+    fn load(asset_server: &AssetServer) -> impl Into<AssetMap<Self>> {
+        [(
+            Self::Ducky,
             asset_server.load_with_settings(
                 "images/ducky.png",
                 |settings: &mut ImageLoaderSettings| {
                     settings.sampler = ImageSampler::nearest();
                 },
             ),
-        )]))
+        )]
     }
 }
 
@@ -53,21 +49,21 @@ pub enum SfxKey {
 impl AssetKey for SfxKey {
     type Asset = AudioSource;
 
-    fn load(asset_server: &AssetServer) -> AssetMap<Self> {
-        AssetMap(HashMap::from([
+    fn load(asset_server: &AssetServer) -> impl Into<AssetMap<Self>> {
+        [
             (
-                SfxKey::ButtonHover,
+                Self::ButtonHover,
                 asset_server.load("audio/sfx/button_hover.ogg"),
             ),
             (
-                SfxKey::ButtonPress,
+                Self::ButtonPress,
                 asset_server.load("audio/sfx/button_press.ogg"),
             ),
-            (SfxKey::Step1, asset_server.load("audio/sfx/step1.ogg")),
-            (SfxKey::Step2, asset_server.load("audio/sfx/step2.ogg")),
-            (SfxKey::Step3, asset_server.load("audio/sfx/step3.ogg")),
-            (SfxKey::Step4, asset_server.load("audio/sfx/step4.ogg")),
-        ]))
+            (Self::Step1, asset_server.load("audio/sfx/step1.ogg")),
+            (Self::Step2, asset_server.load("audio/sfx/step2.ogg")),
+            (Self::Step3, asset_server.load("audio/sfx/step3.ogg")),
+            (Self::Step4, asset_server.load("audio/sfx/step4.ogg")),
+        ]
     }
 }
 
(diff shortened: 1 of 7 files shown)


paul-hansen (approved):
Looks great! I like the generic AssetMap, the name is much better too, the previous naming (e.g. ImageAsset) was confusing to use when we already have `Assets<Image>` as variable names become too similar.
Thanks!

src/game/assets.rs
```diff
@@ -44,58 +50,64 @@ pub enum SfxAsset {
...
+}
+
+#[derive(Resource, Reflect, Deref, DerefMut)]
+#[reflect(Resource)]
+pub struct AssetMap<K: AssetKey>(HashMap<K, Handle<K::Asset>>);
+
+impl<K: AssetKey> AssetMap<K> {
+    pub fn all_loaded(&self, asset_server: &AssetServer) -> bool {
+        self.0
+            .values()
```
janhohenheim, line 110:
```suggestion
        self.values()
```

src/game/assets.rs
```diff
@@ -44,58 +50,64 @@ pub enum SfxAsset {
...
+    type Asset: Asset;
 
-    pub fn all_loaded(&self, assets: &Assets<AudioSource>) -> bool {
-        self.0.iter().all(|(_, handle)| assets.contains(handle))
+    fn load(asset_server: &AssetServer) -> AssetMap<Self>;
+}
+
+#[derive(Resource, Reflect, Deref, DerefMut)]
+#[reflect(Resource)]
+pub struct AssetMap<K: AssetKey>(HashMap<K, Handle<K::Asset>>);
```
janhohenheim, line 105: I'd implement `From<K, V, const N: usize> for AssetMap`, make `load` return `impl Into<AssetMap>` and then make the `load` implementations return the array directly. Less boilerplate that way.

src/game/assets.rs
```diff
@@ -44,58 +50,64 @@ pub enum SfxAsset {
...
+            ),
+            (
+                SoundtrackKey::Gameplay,
+                asset_server.load("audio/soundtracks/Fluffing A Duck.ogg"),
+            ),
+        ]))
     }
+}
+
+pub trait AssetKey: Sized {
```
janhohenheim, line 97: I get why this is called `Key`, but it highlights the implementation detail and not the intention behind it. Also, the function `load` doesn't map semantic sense on a key and implies this is conceptually the asset itself. Maybe something like `PreloadedAsset` would be nicer?

src/game/assets.rs
```diff
@@ -4,38 +4,44 @@ use bevy::{
...
+    app.register_type::<(
+        AssetMap<ImageKey>,
+        AssetMap<SfxKey>,
+        AssetMap<SoundtrackKey>,
+    )>();
+    app.add_systems(Startup, preload_assets);
+}
+
+fn preload_assets(mut commands: Commands, asset_server: Res<AssetServer>) {
+    commands.insert_resource(ImageKey::load(&asset_server));
```
janhohenheim, line 17: Implement `FromWorld` for all `T: AssetKey + Resource` and these can be simple `init_resource` calls.

src/screen/loading.rs
```diff
@@ -5,41 +5,38 @@ use bevy::prelude::*;
...
-    commands.insert_resource(SoundtrackAssets::new(&asset_server));
+fn all_assets_loaded(
+    asset_server: Res<AssetServer>,
+    image_map: Res<AssetMap<ImageKey>>,
+    sfx_map: Res<AssetMap<SfxKey>>,
+    soundtrack_map: Res<AssetMap<SoundtrackKey>>,
+) -> bool {
+    image_map.all_loaded(&asset_server)
+        && sfx_map.all_loaded(&asset_server)
+        && soundtrack_map.all_loaded(&asset_server)
```
janhohenheim, line 37: Would be cool to use polymorphism here to reduce duplication, but I don't think that's possible here, is it? Some thing like

```rust
[image_map as IdkWhichType, sfx_map, soundtrack_map].iter().all(...)
```

src/game/spawn/player.rs
```diff
@@ -26,7 +26,7 @@ pub struct Player;
 fn spawn_player(
     _trigger: Trigger<SpawnPlayer>,
     mut commands: Commands,
-    images: Res<ImageAssets>,
+    image_map: Res<AssetMap<ImageKey>>,
```
janhohenheim, line 29: Given that we also use a sprite atlas in the context of a player, which is also a kind of "image map", this is some potentially very confusing wording. Consider finding a better name for `AssetMap` in general.

janhohenheim (changes requested):

janhohenheim:
Change the design doc accordingly after the review is through, but before merging.

src/game/assets.rs
benfrankel: It'd be `FromWorld` for `AssetMap<K>`, and this would become `app.init_resource::<AssetMap<ImageKey>>()`. Which is reasonable.

src/game/assets.rs
janhohenheim: @paul-hansen do you have an opinion here?

src/game/spawn/player.rs
janhohenheim: @paul-hansen do you have an opinion here?

benfrankel: How about `image_handles: Res<HandleMap<ImageKey>>`?

src/screen/loading.rs
benfrankel: I did think about this. I'm not sure there's a more elegant option than just calling `.all_loaded` on each one, without ballooning the complexity. If there is a way, it should be a separate PR probably.

src/game/assets.rs
janhohenheim: I think that, fundamentally, actual keys are not a thing that loads anything

src/game/spawn/player.rs
janhohenheim: Works for me

src/game/assets.rs
benfrankel: This is going to be a tradeoff between conceptual purity and code quality.

benfrankel: I pushed a potential workaround for this, removing `AssetKey::load` in favor of manual impls of `FromWorld` for `AssetMap<XyzKey>`. It adds some boilerplate for each new asset collection, but you get the conceptual purity.

janhohenheim: My argument stems from the idea that `Key` is a bit of an implementation detail anyways and does not help you conceptualize what is happening, hence `PreloadedAsset`.
I'm fine with leaving it like this if you disagree, especially considering @paul-hansen's endorsement as an unbiased third party.

benfrankel: I'd argue that the preloaded part is the implementation detail that the rest of the codebase doesn't need to know or care about. :P

src/game/assets.rs
```diff
@@ -44,58 +50,78 @@ pub enum SfxAsset {
...
+            (
+                SoundtrackKey::Gameplay,
+                asset_server.load("audio/soundtracks/Fluffing A Duck.ogg"),
+            ),
+        ]
+        .into()
+    }
+}
+
+pub trait AssetKey: Sized {
```
janhohenheim, line 105: I think this trait is no longer needed

janhohenheim (changes requested):

src/game/spawn/player.rs
paul-hansen: No strong opinions here, since it's just a template so we can always rename to whatever we want.
I think making it clearer that it's handles not the actual assets might help some new people with the concepts slightly so I'm on board with that.

src/game/assets.rs
benfrankel: It's needed so that `AssetMap` (or `HandleMap`) only requires one generic.

src/game/spawn/player.rs
benfrankel: Renamed to `HandleMap`.

janhohenheim (changes requested):
Looks good now! Will approve after the design doc is updated

benfrankel:
@janhohenheim Updated design doc.

janhohenheim (approved):

PR #175, opened 2024-07-18
Improve documentation

### LINKED ISSUE #171: Improve documentation of `Spawn`

janhohenheim:
Improves some things that are not documented clearly enough that @christopherbiscardi discovered while using the template.
Resolves #171

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,6 +78,12 @@ fn spawn_camera(mut commands: Commands) {
     commands.spawn((
         Name::new("Camera"),
         Camera2dBundle::default(),
+        // Render all UI to this camera.
+        // Not strictly necessary since we only use one camera,
+        // but if you don't use this component, your UI will disappear as soon
+        // as you add another camera. This includes indirect ways of adding cameras like using
+        // [ui node outlines](https://bevyengine.org/news/bevy-0-14/#ui-node-outline-gizmos)
+        // for debugging. So it's good to have this here for future-proofing.
         IsDefaultUiCamera,
     ));
 }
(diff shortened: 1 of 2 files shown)


src/ui/widgets.rs
```diff
@@ -144,6 +144,9 @@ impl Containers for Commands<'_, '_> {
 }
 
 /// An internal trait for types that can spawn entities.
+/// This is just here so that [`Widgets`] can be implemented on all types that
```
ChristopherBiscardi, line 147:
```suggestion
/// This is here so that [`Widgets`] can be implemented on all types that
```

ChristopherBiscardi:
smol suggestion. Reasoning being the same as others in a similar vein. otherwise 👍🏼 lgtm

src/lib.rs
```diff
@@ -78,6 +78,12 @@ fn spawn_camera(mut commands: Commands) {
     commands.spawn((
         Name::new("Camera"),
         Camera2dBundle::default(),
+        // Render all UI to this camera.
+        // Not strictly necessary since we only use one camera,
+        // but if you don't use this component, your UI will disappear as soon
+        // as you add another camera. This includes indirect ways of adding cameras like using
+        // [ui node outlines](https://bevyengine.org/news/bevy-0-14/#ui-node-outline-gizmos)
+        // for debugging. So it's good to have this here for future-proofing.
```
benfrankel, line 86:
```suggestion
        // Render all UI to this camera.
        // Not strictly necessary since we only use one camera,
        // but if we don't use this component, our UI will disappear as soon
        // as we add another camera. This includes indirect ways of adding cameras like using
        // [ui node outlines](https://bevyengine.org/news/bevy-0-14/#ui-node-outline-gizmos)
        // for debugging. So it's good to have this here for future-proofing.
```

benfrankel (changes requested):

benfrankel (approved):

PR #177, opened 2024-07-18
Refactor audio a bit

benfrankel:
Overview of changes:
- Make some comments a bit more explicit.
- Derive `Copy, Clone` for asset keys.
- Add `sfx::plugin` and `soundtrack::plugin`.
- Rename marker component `SoundtrackMarker` -> `IsSoundtrack`.
- Rename events `Sfx` -> `PlaySfx` and `Soundtrack` -> `PlaySoundtrack`.
- Combine specific-key variants of `PlaySfx` and `PlaySoundtrack`.
- Simplify `play_sfx` and `play_soundtrack` function bodies (this might be controversial).
- Move `game::audio::sfx::button_interaction_sfx` -> `ui::interaction::trigger_interaction_sfx`.

### DIFF
--- a/src/game/audio/sfx.rs
+++ b/src/game/audio/sfx.rs
@@ -3,40 +3,40 @@ use rand::seq::SliceRandom;
 
 use crate::game::assets::{HandleMap, SfxKey};
 
-pub(super) fn play_sfx(
-    trigger: Trigger<Sfx>,
+pub(super) fn plugin(app: &mut App) {
+    app.observe(play_sfx);
+}
+
+fn play_sfx(
+    trigger: Trigger<PlaySfx>,
     mut commands: Commands,
     sfx_handles: Res<HandleMap<SfxKey>>,
 ) {
-    let event = trigger.event();
-    let source = match event {
-        Sfx::ButtonHover => &sfx_handles[&SfxKey::ButtonHover],
-        Sfx::ButtonPress => &sfx_handles[&SfxKey::ButtonPress],
-        Sfx::Step => random_step(&sfx_handles),
-    }
-    .clone_weak();
-    let settings = PlaybackSettings {
-        mode: PlaybackMode::Despawn,
-        ..default()
-    };
-    commands.spawn(AudioSourceBundle { source, settings });
+    commands.spawn(AudioSourceBundle {
+        source: sfx_handles[&match trigger.event() {
+            PlaySfx::ButtonHover => SfxKey::ButtonHover,
+            PlaySfx::ButtonPress => SfxKey::ButtonPress,
+            PlaySfx::Step => random_step(),
+        }]
+            .clone_weak(),
+        settings: PlaybackSettings {
+            mode: PlaybackMode::Despawn,
+            ..default()
+        },
+    });
 }
 
-/// Play a single sound effect.
+/// Trigger this event to play a single sound effect.
 #[derive(Event)]
-pub enum Sfx {
+pub enum PlaySfx {
     ButtonHover,
     ButtonPress,
     Step,
 }
 
-fn random_step(sfx_handles: &HandleMap<SfxKey>) -> &Handle<AudioSource> {
-    [
-        &sfx_handles[&SfxKey::Step1],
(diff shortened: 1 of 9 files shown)


src/game/audio/sfx.rs
```diff
@@ -3,40 +3,40 @@ use rand::seq::SliceRandom;
...
-fn random_step(sfx_handles: &HandleMap<SfxKey>) -> &Handle<AudioSource> {
-    [
-        &sfx_handles[&SfxKey::Step1],
-        &sfx_handles[&SfxKey::Step2],
-        &sfx_handles[&SfxKey::Step3],
-        &sfx_handles[&SfxKey::Step4],
-    ]
-    .choose(&mut rand::thread_rng())
-    .unwrap()
+fn random_step() -> SfxKey {
```
janhohenheim, line 37: You could implement this on `SfxKey` now

src/game/audio/sfx.rs
```diff
@@ -3,40 +3,40 @@ use rand::seq::SliceRandom;
...
-        mode: PlaybackMode::Despawn,
-        ..default()
-    };
-    commands.spawn(AudioSourceBundle { source, settings });
+    commands.spawn(AudioSourceBundle {
+        source: sfx_handles[&match trigger.event() {
+            PlaySfx::ButtonHover => SfxKey::ButtonHover,
+            PlaySfx::ButtonPress => SfxKey::ButtonPress,
+            PlaySfx::Step => random_step(),
+        }]
```
janhohenheim, line 20: Lots happening in a very condensed space. Could you split this into multiple variables and steps?

src/game/audio/soundtrack.rs
```diff
@@ -2,43 +2,49 @@ use bevy::{audio::PlaybackMode, prelude::*};
...
-        mode: PlaybackMode::Loop,
-        ..default()
-    };
-    commands.spawn((AudioSourceBundle { source, settings }, SoundtrackMarker));
+    commands.spawn((
+        AudioSourceBundle {
+            source: soundtrack_handles[&match trigger.event() {
+                PlaySoundtrack::Disable => return,
+                PlaySoundtrack::Credits => SoundtrackKey::Credits,
+                PlaySoundtrack::Gameplay => SoundtrackKey::Gameplay,
```
janhohenheim, line 25: Maybe an `impl TryFrom<Playsoundtrack> for SoundtrackKey` would help?

src/game/audio/soundtrack.rs
```diff
@@ -2,43 +2,49 @@ use bevy::{audio::PlaybackMode, prelude::*};
...
-    .clone_weak();
 
-    let settings = PlaybackSettings {
-        mode: PlaybackMode::Loop,
-        ..default()
-    };
-    commands.spawn((AudioSourceBundle { source, settings }, SoundtrackMarker));
+    commands.spawn((
+        AudioSourceBundle {
+            source: soundtrack_handles[&match trigger.event() {
```
janhohenheim, line 22: Same as above in similar code

janhohenheim (changes requested):
Some minor code quality things, but I like the idea :) 
I cans see a pattern emerging here, so maybe we should add this to the design doc.

src/game/audio/sfx.rs
benfrankel: True. I was thinking about doing the same for `PlaySoundtrack`, but then I realized that it wouldn't be able to `return` from the caller without being a macro, so I didn't do it for either case.

I can see only implementing this on `SfxKey` specifically?

janhohenheim: That would be more of a `impl TryFrom<PlaySoundtrack>` imo so that you can return on `None`

src/game/audio/sfx.rs
benfrankel: Split getting the key out into an earlier step.

src/game/audio/soundtrack.rs
benfrankel: Split getting the key out into an earlier step.

janhohenheim (approved):

PR #181, opened 2024-07-19
Fix missing flags and always enable `bevy/wayland`

benfrankel:


janhohenheim (approved):

benfrankel:
May actually need to mention `--features bevy/wayland` for the dev build too :thinking:

janhohenheim (approved):

PR #214, opened 2024-08-04
Use binstall to install wasm-bindgen and wasm-opt

MeoMix:
Trunk automatically installs prerequisites, if they're not installed, when calling `trunk build --release`. It looks something like this when it occurs:

image

wasm-bindgen installs very quickly, but wasm-opt takes a non-trivial amount of time to install.

It is possible to install these dependencies using `cargo binstall` which can yield time savings.

On my own repository I had a build time of `4m 11s` prior to using `binstall wasm-opt` and a build time of `2m 40s` afterward.
Additionally, I had a build time of `2m 31s` after using `binstall wasm-bindgen-cli`, but I attribute some of this to variance because trunk installing wasm-bindgen only took 5 seconds in the above screenshot.

Nevertheless, it seems worthwhile to install `wasm-opt` via `cargo binstall` and I figured might as well install `wasm-bindgen-cli` as well.

One concern I had is that it's wasm-bindgen*-cli* which seems excessive, but it's what https://github.com/rustwasm/wasm-bindgen suggests... so I went with it.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -122,7 +122,7 @@ jobs:
       - name: Install and run trunk (Web)
         if: ${{ matrix.platform == 'web' }}
         run: |
-          cargo binstall trunk wasm-bindgen-cli wasm-opt --no-confirm
+          cargo binstall --no-confirm trunk wasm-bindgen-cli wasm-opt
           trunk build --release --dist '${{ env.OUT_DIR }}'
 
       - name: Build binaries (non-Web)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -122,7 +122,7 @@ jobs:
       - name: Install and run trunk (Web)
         if: ${{ matrix.platform == 'web' }}
         run: |
-          cargo binstall trunk --no-confirm
+          cargo binstall trunk wasm-bindgen-cli wasm-opt --no-confirm
```
benfrankel, line 125:
```suggestion
          cargo binstall --no-confirm trunk wasm-bindgen-cli wasm-opt
```
Probably nice to move the flag earlier now, but not important.

benfrankel (approved):

benfrankel:
Optimizing the workflows is appreciated :)

janhohenheim:
Thanks a bunch!

Issue #203, opened 2024-07-30
Address Alice's notes


benfrankel:
- [x] **I don't think `InteractionQuery` is worth the cognitive complexity cost**
- [x] **`InteractionPalette`'s naming / docs aren't clear enough** (I like the pattern, but it needs to be taught better)
- [x] ~~**`handle_title_action` should be done using observers, rather than a `TitleAction` component**~~
  - Solved using one-shot systems instead. See the motivation in https://github.com/TheBevyFlock/bevy_quickstart/pull/229
- [x] **`exit_playing` and friends should be renamed to reflect a smaller scope: state transition schedules should contain multiple systems by convention**
- [x] **audio should be data-driven, rather than using enums, and needs a generalized abstraction for "random within this group"**
- [x] **I don't like the layout of the `game` module: IMO trying to further separate out gameplay logic from things like sound effects is important**
- [x] `WrapWithinWindow` is probably clearer as `ScreenWrap`
- [x] `MovementController` needs docs
- [x] ~~you should be able to make `AppSet` more robust to refactors by using the `PartialOrd` trait~~
  - Removed per [this discussion](https://discord.com/channels/691052431525675048/1258521739395203174/1269796709819617443)
- [x] "Alice - Foo" in credits screen will be clearer with a made up job title instead of `Foo`
- [x] "We hold no patent rights to anything presented in this repository." This should be clearer or probably just removed.
- [x] `screen` is kind of a weird name for that module 🤔 Totally just a nit though, and I don't really have better suggestions
- [x] I'm not convinced that spawning things like the player should be an observer; that seems like a pointless layer of indirection over just using a custom command
- [x] man we really need an actual 2D animation API <:pensive_cowboy:791012102655574086> 
- [x] love the Known Issues page: that's a really nice compromise

benfrankel:
A rename option suggested by @alice-i-cecile: `screen` -> `screens`

janhohenheim:
The animation API is resolved too, not much we can do.

MiniaczQ:
Discussion https://discord.com/channels/691052431525675048/1258521739395203174/1268339778240381061

benfrankel:
See https://github.com/TheBevyFlock/bevy_quickstart/issues/223 for the point about spawning.

janhohenheim:
See https://github.com/bevyengine/bevy/pull/14554 for how to do sounds (thanks Alice!)

janhohenheim:
All done now :D

PR #221, opened 2024-08-04
Remove InteractionQuery

janhohenheim:
Part of #203


rparrett (approved):
Beat me to it

benfrankel (approved):
Yeah, this is fine. `bevy_mod_picking` would be the alternative here and then these systems wouldn't even be needed.

PR #222, opened 2024-08-04
Rename enter_* and exit_* systems

janhohenheim:
Part of #203

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -230,11 +230,11 @@ For each screen, create a plugin that handles the setup and teardown of the scre
 ```rust
 // game_over.rs
 pub(super) fn plugin(app: &mut App) {
-    app.add_systems(OnEnter(Screen::GameOver), enter_game_over);
-    app.add_systems(OnExit(Screen::GameOver), exit_game_over);
+    app.add_systems(OnEnter(Screen::GameOver), show_game_over_screen);
+    app.add_systems(OnExit(Screen::GameOver), reset_highscore);
 }
 
-fn enter_game_over(mut commands: Commands) {
+fn show_game_over_screen(mut commands: Commands) {
     commands.
         .ui_root()
         .insert(StateScoped(Screen::GameOver))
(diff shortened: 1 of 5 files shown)


docs/design.md
```diff
@@ -243,9 +243,8 @@ fn enter_game_over(mut commands: Commands) {
         });
 }
 
-fn exit_game_over(mut next_screen: ResMut<NextState<Screen>>) {
-    // Go back to the title screen
-    next_screen.set(Screen::Title);
+fn reset_game_state(mut highscore: ResMut<Highscore>) {
```
benfrankel, line 246:
```suggestion
fn reset_highscore(mut highscore: ResMut<Highscore>) {
```

src/screens/credits.rs
```diff
@@ -25,7 +25,7 @@ enum CreditsAction {
     Back,
 }
 
-fn enter_credits(mut commands: Commands) {
+fn show_credits(mut commands: Commands) {
```
benfrankel, line 28: We should be consistent in naming `show_credits` vs. `show_credits_screen` between the different screens. I'm fine with either.

src/screens/title.rs
```diff
@@ -22,7 +22,7 @@ enum TitleAction {
     Exit,
 }
 
-fn enter_title(mut commands: Commands) {
+fn show_title_menu(mut commands: Commands) {
```
benfrankel, line 25:
```suggestion
fn show_title_screen(mut commands: Commands) {
```

Or just `show_title`, depending on the choice for consistency.

benfrankel (changes requested):

src/screens/title.rs
janhohenheim: Disagree. There may be multiple kinds of title screen. Depending on your view, the splash may also be part of the title. This is specifically the title screen *menu*

src/screens/credits.rs
janhohenheim: [`f465e65` (#222)](https://github.com/TheBevyFlock/bevy_quickstart/pull/222/commits/f465e65a1fb8f50e06dbdf9a735869ae71f4ef8c)

src/screens/title.rs
benfrankel: The splash isn't part of the title screen though, because we explicitly have `Screen::Splash` distinct from `Screen::Title`. I've also never heard the term "title menu", only "title screen" and "main menu"

benfrankel (approved):

PR #225, opened 2024-08-05
Use data-driven sound effects

janhohenheim:
Part of #203 
Adapted from https://github.com/bevyengine/bevy/pull/14554
I'll update the design doc in a follow-up.

~~I don't like how some assets use strictly `enum`s as identifiers and one uses `AsRef<str>`. It seems like they should all use the same kind of key. The thing is, I'm still not entirely convinced by the `String` key. I need some more opinions on this.~~
All assets now use `String`s as keys for consistency.


janhohenheim:
Ping @alice-i-cecile for feedback

alice-i-cecile:
For this template, I think the string-based approach is better.

janhohenheim:
@alice-i-cecile does that mean you suggest using backing strings for every asset kind and use enum accessors with `AsRef<str>` impls for them all? Or would you use string constants for all keys instead?

janhohenheim:
I went ahead and removed the `enum` in favor of constants. Should I do the same for the other assets?

src/game/assets.rs
```diff
@@ -5,123 +5,105 @@ use bevy::{
...
 
-pub trait AssetKey: Sized {
-    type Asset: Asset;
-}
-
-#[derive(Resource, Reflect, Deref, DerefMut)]
+/// This uses a map of `String` to `Vec<Handle<AudioSource>>` to store sound effects instead
+/// of the `HandleMap` used for the other assets.
+/// The key is a `String` because sound effects are often tied closely to file paths or other strings.
+/// The value is a `Vec<Handle<AudioSource>>` because a single sound effect can have multiple variations.
```
benfrankel, line 77: Outdated comment. There is no `HandleMap` and everything else is using `String` as a key as well, so explaining that choice here specifically doesn't make sense. The only thing that needs to be explained is that the `Vec<_>` is for variations.

benfrankel (changes requested):
Still reviewing. Noting that `XyzHandles::FOO` reads wrong because it's the key, not the handle.

src/game/audio/sound_effects.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn apply(self, world: &mut World) {
+        // Access both the world and the resource we need from it using resource_scope
+        // which temporarily removes the SoundEffects resource from the world
+        world.resource_scope(|world, mut sound_effects: Mut<SoundEffectHandles>| {
+            sound_effects.play(self.name, world, self.settings);
+        });
+    }
+}
+
+/// An "extension trait" used to make it convenient to play sound effects via [`Commands`].
```
benfrankel, line 65:
```suggestion
/// An extension trait used to make it convenient to play sound effects via [`Commands`].
```
This really is an extension trait :smile:

src/game/audio/sound_effects.rs
```diff
@@ -0,0 +1,100 @@
...
+///
+/// This technique allows us to implement methods for types that we don't own,
+/// which can be used as long as the trait is in scope.
```
benfrankel, line 68: Redundant, the "via [\`Commands\`]" above already explains this.

src/game/assets.rs
janhohenheim: [`212dd2e` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/212dd2ea61bc0fd3c64e5022def83bdd96d5ed6c)

janhohenheim:
Added a `KEY_` prefix to all keys: [`e477525` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/e4775259ac9b41a4ea4e6457a4ace2213b5d40a7)

src/game/audio/sound_effects.rs
benfrankel, line 1: The comments in here read like an example rather than the user's own code.

src/game/audio/sound_effects.rs
```diff
@@ -0,0 +1,100 @@
...
+//! This pattern is adapted from the [Bevy example for sound effects](https://github.com/bevyengine/bevy/pull/14554).
+
+use bevy::{ecs::world::Command, prelude::*};
+use rand::{distributions::Uniform, Rng as _};
+
+use crate::game::assets::SoundEffectHandles;
+
+pub(super) fn plugin(_app: &mut App) {
+    // No setup required for this plugin.
+    // It's still good to have a function here so that you can add some setup later if needed.
```
benfrankel, line 14: I don't agree with this personally.

src/game/audio/soundtrack.rs
```diff
@@ -1,48 +1,68 @@
...
+    fn play_soundtrack(&mut self, name: impl Into<String>);
+
+    fn stop_current_soundtrack(&mut self);
 }
 
-/// Marker component for the soundtrack entity so we can find it later.
-#[derive(Component, Reflect)]
-#[reflect(Component)]
-struct IsSoundtrack;
+impl<'w, 's> SoundtrackCommands for Commands<'w, 's> {
```
benfrankel, line 60:
```suggestion
impl SoundtrackCommands for Commands<'_, '_> {
```
:)

src/screens/loading.rs
```diff
@@ -28,15 +28,34 @@ fn show_loading_screen(mut commands: Commands) {
...
-        && sfx_handles.all_loaded(&asset_server)
+        && sound_effects_handles.all_loaded(&asset_server)
         && soundtrack_handles.all_loaded(&asset_server)
 }
 
 fn continue_to_title(mut next_screen: ResMut<NextState<Screen>>) {
     next_screen.set(Screen::Title);
 }
+
+trait AssetMap {
```
benfrankel, line 44: Would add a brief doc comment here even though it's not `pub`.
```suggestion
/// An extension trait to check if all the assets in an asset collection are loaded.
trait AssetMap {
```

benfrankel (changes requested):
Some more comments.

src/game/audio/sound_effects.rs
janhohenheim: Taken from https://github.com/bevyengine/bevy/pull/14554

src/game/audio/sound_effects.rs
janhohenheim: Taken from https://github.com/bevyengine/bevy/pull/14554

src/game/audio/sound_effects.rs
janhohenheim: Well, that's because they come from an example :D I'll change them, then

src/game/audio/sound_effects.rs
benfrankel: Mmm yeah but that's an example, not a template. I see why the comments read like an example now :smile:

The comments here should probably not match the example 1-1 unless we want to link to the example and make it explicit that that's where this is taken from.

src/game/audio/soundtrack.rs
janhohenheim: Ah, thanks! Added that to the upstream PR as well

src/game/audio/sound_effects.rs
benfrankel: Ok well there is a link in the module-level docs. If it's actually 1-1 I'd change "adapted from" to "taken from".

src/game/audio/sound_effects.rs
janhohenheim: [`cabeee5` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/cabeee5d8e6ea3f350cab88cda624961be500ce8)

src/game/audio/sound_effects.rs
janhohenheim: Fixed in [`cabeee5` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/cabeee5d8e6ea3f350cab88cda624961be500ce8), but you should add this review comment there as well.

src/game/audio/sound_effects.rs
janhohenheim: Fixed in [`cabeee5` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/cabeee5d8e6ea3f350cab88cda624961be500ce8), but you should add this review comment there as well.

janhohenheim: Not *quite* 1-1, but close enough. Changed "adapted from" to "taken from" in [`9452ee2` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/9452ee2aebd5d836adf283059ab2746f6f76a956)

benfrankel: Yeah I'll want to do a separate review of the example PR.

src/game/audio/sound_effects.rs
janhohenheim: I think not adding a plugin when nearly all other files have a plugin is potentially more confusing to a user than the nuance is worth. I personally would add a `plugin` here because it is very realistic that a `register_type` or `observe` might one day get in here, and I don't want to be annoyed that I have to go back and wire the plugin up. One plugin per file is a good rule of thumb imo.

src/ui/interaction.rs
```diff
@@ -65,8 +65,12 @@ fn trigger_interaction_sfx(
 ) {
     for interaction in &interactions {
         match interaction {
-            Interaction::Hovered => commands.trigger(PlaySfx::Key(SfxKey::ButtonHover)),
-            Interaction::Pressed => commands.trigger(PlaySfx::Key(SfxKey::ButtonPress)),
+            Interaction::Hovered => {
+                commands.play_sound_effect(SoundEffectHandles::KEY_BUTTON_HOVER)
```
benfrankel, line 69: I think it's interesting that we defined an extension trait and this is still more verbose than it was. Mostly due to spelling out `sound_effect` every time instead of `sfx`. Not necessarily a suggestion, just a remark :p

benfrankel (changes requested):

src/ui/interaction.rs
janhohenheim: That's true, hehe.

src/game/audio/sound_effects.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// while string keys are a better fit for sound effects corresponding to objects loaded from a data file.
+    fn play(&mut self, name: impl AsRef<str>, world: &mut World, settings: PlaybackSettings) {
+        let name = name.as_ref();
+        if let Some(sfx_list) = self.get_mut(name) {
+            // If we need precise control over the randomization order of our sound effects,
+            // we can store the RNG as a resource and modify these functions to take it as an argument.
+            let rng = &mut rand::thread_rng();
+            let random_sfx = sfx_list.choose(rng).unwrap();
+
+            // We don't need a (slightly) more expensive strong handle here (which are used to keep assets loaded in memory)
```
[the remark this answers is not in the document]
janhohenheim: Accepted, but add this to https://github.com/bevyengine/bevy/pull/14554 as well

src/screens/loading.rs
```diff
@@ -28,15 +28,35 @@ fn show_loading_screen(mut commands: Commands) {
...
+        && sound_effects_handles.all_loaded(&asset_server)
         && soundtrack_handles.all_loaded(&asset_server)
 }
 
 fn continue_to_title(mut next_screen: ResMut<NextState<Screen>>) {
     next_screen.set(Screen::Title);
 }
+
+/// An extension trait to check if all the assets in an asset collection are loaded.
+trait AssetMap {
```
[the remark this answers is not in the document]
janhohenheim: You're absolutely right: [`fc7f665` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/fc7f665154c2d5c2bf94b5fd9bca2547a4b79aa7)

src/game/audio/sound_effects.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl Command for PlaySoundEffect {
+    fn apply(self, world: &mut World) {
+        world.resource_scope(|world, mut sound_effects: Mut<SoundEffectHandles>| {
+            sound_effects.play(self.name, world, self.settings);
+        });
+    }
+}
+
+/// An extension trait used to make it convenient to play sound effects via [`Commands`].
```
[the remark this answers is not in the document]
janhohenheim: Used your version. Suggest this to https://github.com/bevyengine/bevy/pull/14554 as well.

src/game/audio/sound_effects.rs
benfrankel: Not strongly enough to block this PR. Can always discuss this later.

src/game/audio/sound_effects.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// When defining the settings for this method, we almost always want to use [`PlaybackMode::Despawn`](bevy::audio::PlaybackMode).
+    /// Every time a sound effect is played, a new entity is generated. Once the sound effect is complete,
+    /// the entity should be cleaned up, rather than looping or sitting around uselessly.
+    ///
+    /// This method accepts any type which implements `AsRef<str>`.
+    /// This allows us to pass in `&str`, `String`, or a custom type that can be converted to a string.
+    ///
+    /// These custom types can be useful for defining enums that represent specific sound effects.
+    /// Generally speaking, enum values should be used to represent one-off or special-cased sound effects,
+    /// while string keys are a better fit for sound effects corresponding to objects loaded from a data file.
```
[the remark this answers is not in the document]
janhohenheim: [`e968cfb` (#225)](https://github.com/TheBevyFlock/bevy_quickstart/pull/225/commits/e968cfb28c502426134d85ec76b89fecc166e098)

benfrankel (approved):

Issue #223, opened 2024-08-04
Spawn entities with an `EntityCommand`-based abstraction


benfrankel:
An example of how the proposed API looks in practice: https://github.com/benfrankel/bevy_jam_template/blob/main/src/screen/title.rs.

Benefits:
- You get an `EntityCommands` back immediately.
- You can access `SystemParam`s ergonomically by adding them as arguments.
- You can pass in extra data (e.g. `struct Projectile { lifetime, .. }`).
- Converting a "pure" entity spawning system to one that takes in extra data is relatively easy:
    - Define your data type, like `struct Projectile`.
    - Implement `EntityCommand` for `Projectile`, and delegate to the spawning system with `world.run_system_once_with((id, self), projectile)`.
    - Change `In(id): In<Entity>` to `In((id, this)): In<(Entity, Projectile)>`.

Drawbacks:
- Requires extension traits.
- Unfamiliar approach (using observers to spawn things is unfamiliar as well).
- `spawn_fn` and `spawn_with` unfortunately couldn't be merged into one method.
- A few lines of boilerplate to take in data. This could use a VS Code snippet to make it easier.

janhohenheim:
I'm in favor.
Note that we went through many designs and this seems the cleanest by far imo. Bevy does not have the required answer to "How do I spawn an object composed of lots of stuff?" yet while still returning an `EntityCommands`, so we *need* to introduce an abstraction here.

alice-i-cecile:
I'm also in favour of this.

janhohenheim:
Blocked on upstreaming https://github.com/bevyengine/bevy/issues/14278 and maybe https://github.com/bevyengine/bevy/issues/14231

benfrankel:
Closing this since it's definitely not going to happen in the form originally proposed. We've moved on to Bevy 0.16 spawning API, which still has issues but should hopefully improve further in the coming releases.

PR #229, opened 2024-08-05
Add `OnPress` callbacks

janhohenheim:
Pattern taken from @Freyja-moth. Thanks Freyja!

Alternative to the following:
> `handle_title_action` should be done using observers, rather than a `TitleAction` component

[Relevant discussion](https://discord.com/channels/691052431525675048/1258521739395203174/1270111179494133840).  
The TL;DR is that registering a similar `OnPress<T: Event>(pub T)`, which would be need to trigger observers on presses, would require you to register your `T` in your menu in order to register the actual generic system for your `T`. Which seems like a weird extra abstraction to add and a potential source of errors, same a when a user forgets to do `add_event` for `EventReader`s and `EventWriter`s.
That, or you setup your logic for triggering observers on presses separately for every menu, which seems like a lot of boilerplate for something as trivial as pressing on a button.
Additionally, adding observer structs introduces additional indirection where none is needed.

The ugly reflection stuff is due to https://github.com/bevyengine/bevy/issues/14496.
The need for one-shot systems to be registered at the top of the function will be gone after https://github.com/TheBevyFlock/bevy_quickstart/issues/223

### DIFF
--- a/src/screens/credits.rs
+++ b/src/screens/credits.rs
@@ -5,7 +5,7 @@ use bevy::prelude::*;
 use super::Screen;
 use crate::{
     game::{assets::SoundtrackKey, audio::soundtrack::PlaySoundtrack},
-    ui::{prelude::*},
+    ui::prelude::*,
 };
 
 pub(super) fn plugin(app: &mut App) {
@@ -14,7 +14,7 @@ pub(super) fn plugin(app: &mut App) {
 }
 
 fn show_credits_screen(mut commands: Commands) {
-    let on_back = commands.register_one_shot_system(on_back);
+    let enter_title = commands.register_one_shot_system(enter_title);
 
     commands
         .ui_root()
@@ -29,7 +29,7 @@ fn show_credits_screen(mut commands: Commands) {
             children.label("Ducky sprite - CC0 by Caz Creates Games");
             children.label("Music - CC BY 3.0 by Kevin MacLeod");
 
-            children.button("Back").insert(OnPress(on_back));
+            children.button("Back").insert(OnPress(enter_title));
         });
 
     commands.trigger(PlaySoundtrack::Key(SoundtrackKey::Credits));
@@ -39,6 +39,6 @@ fn disable_soundtrack(mut commands: Commands) {
     commands.trigger(PlaySoundtrack::Disable);
 }
 
-fn on_back(mut next_screen: ResMut<NextState<Screen>>) {
+fn enter_title(mut next_screen: ResMut<NextState<Screen>>) {
     next_screen.set(Screen::Title);
 }
--- a/src/screens/credits.rs
+++ b/src/screens/credits.rs
@@ -5,7 +5,7 @@ use bevy::prelude::*;
 use super::Screen;
 use crate::{
     game::{assets::SoundtrackKey, audio::soundtrack::PlaySoundtrack},
-    ui::{interaction::OnPress, prelude::*},
+    ui::{prelude::*},
 };
 
 pub(super) fn plugin(app: &mut App) {
--- a/src/screens/credits.rs
+++ b/src/screens/credits.rs
@@ -5,27 +5,17 @@ use bevy::prelude::*;
 use super::Screen;
 use crate::{
     game::{assets::SoundtrackKey, audio::soundtrack::PlaySoundtrack},
-    ui::prelude::*,
+    ui::{interaction::OnPress, prelude::*},
 };
 
 pub(super) fn plugin(app: &mut App) {
     app.add_systems(OnEnter(Screen::Credits), show_credits_screen);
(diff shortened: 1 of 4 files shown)


janhohenheim:
I also want some feedback from @alice-i-cecile before merging since this is an alternative to her suggestion

benfrankel:
I like `SystemId` here because the API is a lot closer to `bevy_mod_picking`, and the performance impact should be negligible for menu buttons.

janhohenheim:
Note that a possible extension of this could be an API like `On::<Interaction::Press::run(system)`. I would not do that for the template, since it would be trivial to add for a user themselves and I don't think anything outside of `Press` will require a callback. Just wanted to point out the possibility.

alice-i-cecile:
IMO this is more likely to end up using observers, but this is fine until we get a first party solution.

janhohenheim:
@alice-i-cecile do you mean by that that a player will end up changing it to observers or that a first party solution will use observers?

src/screens/credits.rs
```diff
@@ -5,27 +5,17 @@ use bevy::prelude::*;
 use super::Screen;
 use crate::{
     game::{assets::SoundtrackKey, audio::soundtrack::PlaySoundtrack},
-    ui::prelude::*,
+    ui::{interaction::OnPress, prelude::*},
```
benfrankel, line 8: I'd add `OnPress` to the `ui` prelude.

src/screens/title.rs
```diff
@@ -3,54 +3,38 @@
...
-    Credits,
-    /// Exit doesn't work well with embedded applications.
-    #[cfg(not(target_family = "wasm"))]
-    Exit,
 }
 
 fn show_title_screen(mut commands: Commands) {
+    let on_play = commands.register_one_shot_system(on_play);
+    let on_credits = commands.register_one_shot_system(on_credits);
+    let on_exit = commands.register_one_shot_system(on_exit);
```
benfrankel, line 15:
```suggestion
    #[cfg(not(target_family = "wasm"))]
    let on_exit = commands.register_one_shot_system(on_exit);
```
To avoid a clippy warning.

src/ui/interaction.rs
```diff
@@ -17,6 +25,13 @@ pub struct InteractionPalette {
     pub pressed: Color,
 }
 
+/// Component that calls a [one-shot system](https://bevyengine.org/news/bevy-0-12/#one-shot-systems)
+/// when the [`Interaction`] component on the same entity changes to [`Interaction::Pressed`].
+/// Use this in conjuction with [`Commands::register_one_shot_system`] to create a callback for e.g. a button press.
+#[derive(Component, Debug, Reflect, Deref, DerefMut)]
+#[reflect(Component, from_reflect = false)]
+pub struct OnPress(#[reflect(ignore)] pub SystemId);
+
```
benfrankel, line 34: IMO it's nice to have the related system immediately after the component, so I'd move this to right after `apply_interaction_palette` and move `apply_interaction_callback` up to right before `trigger_interaction_sfx`.

Also, IMO rename `apply_interaction_callback` -> `apply_on_press`, since that's the only interaction callback we have atm.

benfrankel (changes requested):
IMO `on_credits` etc. names should instead be named based on the action they perform, like `enter_credits` or `exit`.

src/screens/credits.rs
janhohenheim: [`c81067a` (#229)](https://github.com/TheBevyFlock/bevy_quickstart/pull/229/commits/c81067a2cfe7a7faef579bfa319c75f65d696aeb)

src/ui/interaction.rs
janhohenheim: [`f78efe9` (#229)](https://github.com/TheBevyFlock/bevy_quickstart/pull/229/commits/f78efe9a6337539ae78e1d6448f4794650aeda9c)

Moved system up instead to keep `pub` items above private ones.

benfrankel: I see, I didn't realize we were doing that. Well, as long as it's consistent.

src/ui/interaction.rs
```diff
@@ -17,6 +25,24 @@ pub struct InteractionPalette {
...
+fn apply_on_press(
+    interactions: Query<(&Interaction, &OnPress), Changed<Interaction>>,
+    mut commands: Commands,
+) {
+    for (interaction, OnPress(system_id)) in &interactions {
+        if matches!(interaction, Interaction::Pressed) {
+            commands.run_system(*system_id);
```
benfrankel, line 41:
```suggestion
    for (interaction, OnPress(&system_id)) in &interactions {
        if matches!(interaction, Interaction::Pressed) {
            commands.run_system(system_id);
```
Fine either way.

benfrankel (approved):

benfrankel (approved):

PR #230, opened 2024-08-05
Refactor and document `MovementController`

janhohenheim:

### DIFF
--- a/src/game/movement.rs
+++ b/src/game/movement.rs
@@ -20,7 +20,7 @@ pub(super) fn plugin(app: &mut App) {
     // Record directional input as movement controls.
     app.add_systems(
         Update,
-        record_movement_controller.in_set(AppSet::RecordInput),
+        record_player_directional_input.in_set(AppSet::RecordInput),
     );
 
     // Apply movement based on controls.
@@ -57,7 +57,7 @@ impl Default for MovementController {
     }
 }
 
-fn record_movement_controller(
+fn record_player_directional_input(
     input: Res<ButtonInput<KeyCode>>,
     mut controller_query: Query<&mut MovementController>,
 ) {
--- a/src/game/movement.rs
+++ b/src/game/movement.rs
@@ -34,6 +34,7 @@ pub(super) fn plugin(app: &mut App) {
 
 /// This is the input for our character controller,
 /// which in this case is only the direction the character wants to move in.
+/// For now, this is only used for a single player, but it could power NPCs or other players as well.
 #[derive(Component, Reflect, Default)]
 #[reflect(Component)]
 pub struct MovementController {
--- a/src/game/movement.rs
+++ b/src/game/movement.rs
@@ -33,14 +33,10 @@ pub(super) fn plugin(app: &mut App) {
 /// - Set the movement intent based on directional input.
 /// - Apply movement based on the intent.
 /// - Wrap the character within the window.
-///
-/// We are using a [`Dir2`] here instead of a [`Vec2`] to ensure that the direction is always normalized.
-/// This is important because otherwise the player could move faster diagonally.
-/// If the player does not want to move, the direction is `None`.
 #[derive(Component, Reflect, Default)]
 #[reflect(Component)]
 pub struct MovementController {
-    pub intent: Option<Dir2>,
+    pub intent: Vec2,
 }
 
 fn record_movement_controller(
@@ -62,9 +58,14 @@ fn record_movement_controller(
         intent.x += 1.0;
     }
 
+    // Normalize so that diagonal movement has the same speed as
+    // horizontal and vertical movement.
+    // Skip this if we want to support fractional input from analog sticks.
+    let intent = intent.normalize_or_zero();
+
     // Apply movement intent to controllers.
     for mut controller in &mut controller_query {
-        controller.intent = intent.try_into().ok();
+        controller.intent = intent;
     }
 }
 
@@ -83,10 +84,8 @@ fn apply_movement(
     mut movement_query: Query<(&MovementController, &Movement, &mut Transform)>,
 ) {
     for (controller, movement, mut transform) in &mut movement_query {
-        if let Some(dir) = controller.intent {
-            let velocity = movement.speed * dir;
-            transform.translation += velocity.extend(0.0) * time.delta_seconds();
-        }
+        let velocity = movement.speed * controller.intent;
+        transform.translation += velocity.extend(0.0) * time.delta_seconds();
     }
 }
 
--- a/src/game/movement.rs
+++ b/src/game/movement.rs
@@ -25,9 +25,23 @@ pub(super) fn plugin(app: &mut App) {
     );
 }
 
+/// This is the input for the player's character controller,
+/// which in this case is only the direction the player wants to move in.
+///
+/// A character controller is the collection of systems that govern movement of characters.
+/// In our case, the character controller has the following logic:
+/// - Set the movement intent based on directional input.
+/// - Apply movement based on the intent.
+/// - Wrap the character within the window.
+///
+/// We are using a [`Dir2`] here instead of a [`Vec2`] to ensure that the direction is always normalized.
+/// This is important because otherwise the player could move faster diagonally.
+/// If the player does not want to move, the direction is `None`.
 #[derive(Component, Reflect, Default)]
 #[reflect(Component)]
-pub struct MovementController(pub Vec2);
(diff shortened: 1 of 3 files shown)


src/game/movement.rs
```diff
@@ -25,9 +25,19 @@ pub(super) fn plugin(app: &mut App) {
...
 }
 
+/// This is the input for the player's character controller,
+/// which in this case is only the direction the player wants to move in.
+///
+/// A character controller is the collection of systems that govern movement of characters.
+/// In our case, the character controller has the following logic:
+/// - Set the movement intent based on directional input.
+/// - Apply movement based on the intent.
+/// - Wrap the character within the window.
```
benfrankel, line 35: Hmmm this seems more like module-level documentation. This component doesn't do all of those things.

Also, while this is technically only used as the player's character controller, it doesn't have to be strictly coupled to the player. Though it does currently happen to be, because the `record_movement_controller` system doesn't check for an additional marker component like `KeyboardMovementInput` or whatever.

benfrankel (changes requested):

src/game/movement.rs
```diff
@@ -1,5 +1,12 @@
...
-//! Note that the approach used here is simple for demonstration purposes.
+//! Handle player input and translate it into movement through a character controller.
+/// A character controller is the collection of systems that govern the movement of characters.
+//!
+/// In our case, the character controller has the following logic:
+/// - Set the movement intent based on directional input.
+/// - Apply movement based on the intent.
+/// - Wrap the character within the window.
+///
+//! Note that the implementation used here is simple for demonstration purposes.
```
benfrankel, line 9:
```suggestion
//! Note that the implementation used here is limited for demonstration purposes.
```
To avoid the "simple => you should understand it easily" misunderstanding.

src/game/movement.rs
janhohenheim: [`6877f00` (#230)](https://github.com/TheBevyFlock/bevy_quickstart/pull/230/commits/6877f00864851ed3edf5394c27efe8c143e79c66)

janhohenheim: (And subsequent)

src/game/movement.rs
```diff
@@ -25,9 +32,13 @@ pub(super) fn plugin(app: &mut App) {
     );
 }
 
+/// This is the input for our character controller,
+/// which in this case is only the direction the character wants to move in.
```
benfrankel, line 36:
```suggestion
/// This component acts as a virtual joystick for the character's movement.
```
Not necessarily this wording, but I want to make it clear that this component doesn't represent _input_, like keyboard / mouse input. It's specifically an intermediate layer that represents how the character is trying to move, and it can be populated in any way: keyboard / mouse input, analog stick input, enemy AI, some special system, etc.

src/game/movement.rs
```diff
@@ -1,5 +1,12 @@
-//! Handle player input and translate it into movement.
-//! Note that the approach used here is simple for demonstration purposes.
+//! Handle player input and translate it into movement through a character controller.
+/// A character controller is the collection of systems that govern the movement of characters.
+//!
+/// In our case, the character controller has the following logic:
+/// - Set the movement intent based on directional input.
+/// - Apply movement based on the intent.
+/// - Wrap the character within the window.
+///
```
benfrankel:
```suggestion
//! In our case, the character controller has the following logic:
//! - Set [`MovementController`] based on directional keyboard input.
//! - Apply movement based on [`Movement`] parameters and [`MovementController`] direction.
//! - Wrap the character within the window.
//!
```

benfrankel (changes requested):
Also thinking maybe the `Movement` component should have a short doc comment as well.

src/game/movement.rs
janhohenheim: [`08bf7a9` (#230)](https://github.com/TheBevyFlock/bevy_quickstart/pull/230/commits/08bf7a9af666b5cb1e26ded050e434cb86bc41bf)

src/game/movement.rs
janhohenheim: [`4771b68` (#230)](https://github.com/TheBevyFlock/bevy_quickstart/pull/230/commits/4771b68b43977c73fedeacb47ac210c3d073effb)

src/game/movement.rs
```diff
@@ -1,5 +1,12 @@
-//! Handle player input and translate it into movement.
-//! Note that the approach used here is simple for demonstration purposes.
+//! Handle player input and translate it into movement through a character controller.
+//! A character controller is the collection of systems that govern the movement of characters.
+//!
+//! In our case, the character controller has the following logic:
+//! - Set [`MovementController`] directional intent based on directional keyboard input.
+//! - Apply movement based [`MovementController`] directional intent and maximum speed.
```
benfrankel, line 6:
```suggestion
//! - Set [`MovementController`] intent based on directional keyboard input.
//! - Apply movement based on [`MovementController`] intent and maximum speed.
```
"Directional keyboard input" here means like arrow keys / WASD.

src/game/movement.rs
```diff
@@ -25,9 +32,30 @@ pub(super) fn plugin(app: &mut App) {
...
+    fn default() -> Self {
+        Self {
+            intent: Vec2::ZERO,
+            // 400 pixels per second is a nice default, but we can still vary this per character.
+            max_speed: 400.0,
+        }
+    }
+}
 
 fn record_movement_controller(
```
benfrankel, line 60: I'm kinda feeling like we should rename this system. How about `record_directional_input` or `record_keyboard_movement_input`? Since the input is what we're recording, and we're recording it _into_ the movement controller. But there may be other systems that can record other things into the movement controller.

src/game/movement.rs
```diff
@@ -50,30 +78,21 @@ fn record_movement_controller(
 
     // Normalize so that diagonal movement has the same speed as
     // horizontal and vertical movement.
+    // Skip this if we want to support fractional input from analog sticks.
```
benfrankel, line 81:
```suggestion
    // This should be omitted if the input comes from an analog stick instead.
```

src/game/movement.rs
```diff
@@ -25,9 +32,30 @@ pub(super) fn plugin(app: &mut App) {
...
 #[reflect(Component)]
-pub struct MovementController(pub Vec2);
+pub struct MovementController {
+    /// The direction the character wants to move in.
+    pub intent: Vec2,
+
+    /// Since Bevy's default 2D camera setup is scaled such that
+    /// one unit is one pixel, you can think of this as
+    /// "How many pixels per second should the player move?"
+    /// Note that physics engines may use different unit/pixel ratios.
```
benfrankel, line 46: I feel like this is a little too far in the style of a comment that you'd find in an example, not a template where this will become the user's own code. The same can be said to a smaller degree for some of the other comments.
```suggestion
    /// Maximum speed in world units per second.
```
This is what I'd write in my own code. Maybe a balance can be struck to include a little more explanation, but still worded as though it's the user's own code.

benfrankel (changes requested):

src/game/movement.rs
janhohenheim: I strongly feel like the documentation should stay like this, as this is a question that comes up in Discord all the time. That, or people don't know what a "world unit" is and just try `5.0`, are confused why everything is extremely slow moving.

benfrankel: That's fine, it's mainly the tone / perspective that I'm concerned about.
```suggestion
    /// Maximum speed in world units per second.
    /// 
    /// 1 world unit = 1 pixel with the default camera zoom.
```
for example

src/game/movement.rs
janhohenheim: [`929a7a1` (#230)](https://github.com/TheBevyFlock/bevy_quickstart/pull/230/commits/929a7a1886a0fa8bea889bdaaed6ea0bd627375d)

src/game/movement.rs
janhohenheim: [`70ce354` (#230)](https://github.com/TheBevyFlock/bevy_quickstart/pull/230/commits/70ce354bdbe1509973175176c767fd4d3248393b)

src/game/movement.rs
benfrankel: If I'm being super pedantic, it's not tied to the player. But since it's using keyboard input, which is _probably_ meant to control the player entity, it basically is. So it's fine.

benfrankel (approved):

PR #231, opened 2024-08-06
Replace entity spawning abstraction

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/223.
Part of https://github.com/TheBevyFlock/bevy_quickstart/issues/203.

Will need to update the design doc as well.


MiniaczQ (approved):
pattern looks good, i can look for nits tomorrow if this is still up

janhohenheim (changes requested):
I'm unhappy about the definition site of the construction functions, as I think this kind of function will be very surprising to a newcomer. The solution probably lies in finding a good naming convention, but I can't think of one.

src/screens/playing.rs
```diff
@@ -18,12 +27,27 @@ pub(super) fn plugin(app: &mut App) {
     );
 }
 
+fn playing_screen(In(id): In<Entity>, mut commands: Commands) {
+    commands
+        .entity(id)
+        .add_fn(widget::ui_root)
+        .insert((Name::new("Playing screen"), StateScoped(Screen::Playing)))
+        .with_children(|_children| {
+            // TODO: Spawn playing screen UI here.
```
janhohenheim, line 36:
```suggestion
            // Spawn our playing screen UI like HUDs here.
```

`TODO` sounds like something we forgot to add to the template.

src/ui/widget.rs
```diff
@@ -0,0 +1,206 @@
...
+        self.spawn_with(Header::new(text))
+    }
+
+    fn label(&mut self, text: impl Into<String>) -> EntityCommands {
+        self.spawn_with(Label::new(text))
+    }
+}
+
+/// A simple button with text.
+pub struct Button {
```
janhohenheim, line 39: Maybe rename this to `ButtonWidget` to avoid naming conflicts with `Bevy`.
Also, does this need to be `pub`?

src/spawn.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// An extension trait that provides helper functions for deferred entity spawning.
+pub trait SpawnExt {
+    // Workaround for https://github.com/bevyengine/bevy/issues/14231#issuecomment-2216321086.
+    /// Spawn an entity with an [`EntityCommand`].
+    fn spawn_with<M: 'static>(&mut self, command: impl EntityCommand<M>) -> EntityCommands;
+
+    /// Spawn an entity with a [`System`] that receives the new entity ID via [`In<Entity>`].
+    fn spawn_fn<M>(
```
janhohenheim, line 26: The distinction between `spawn_with` and `spawn_fn` is currently probably impossible to grasp from the naming alone. I suggest renaming this to `insert_with`.

src/spawn.rs
```diff
@@ -0,0 +1,168 @@
...
+            world.run_system_once_with(id, system);
+        });
+        self
+    }
+}
+
+/// An extension trait that allows entity spawning systems to be used as actual systems.
+pub trait SpawnSystemExt<M> {
+    /// Create a system that spawns an entity.
+    fn spawn(self) -> impl Fn(Commands);
```
janhohenheim, line 159: Why can't this already have the signature `fn spawn(commands: Commands)`? We don't need to be object safe.

src/screens/playing.rs
```diff
@@ -18,12 +27,27 @@ pub(super) fn plugin(app: &mut App) {
     );
 }
 
+fn playing_screen(In(id): In<Entity>, mut commands: Commands) {
+    commands
+        .entity(id)
+        .add_fn(widget::ui_root)
```
janhohenheim, line 33: It seems very confusing from a first glance when one should use `add_fn` and when `spawn_fn`. If this is confusing to me as a reviewer, I imagine it would be really really confusing to a newcomer.

janhohenheim (changes requested):

src/ui/widget.rs
benfrankel: The naming conflict shouldn't be an issue since the intended use is `widget::Button`. Unless it's considered that it's an issue regardless. In my own template I have this as `widget::MenuButton` to avoid the conflict.

It needs to be `pub` because you might not be using the `Widgets` extension trait, which only provides spawning functionality, not adding functionality (which you could use to "inherit" from `Button`).

src/screens/credits.rs
```diff
@@ -5,20 +5,23 @@ use bevy::prelude::*;
...
+    spawn::prelude::*,
     ui::prelude::*,
 };
 
 pub(super) fn plugin(app: &mut App) {
-    app.add_systems(OnEnter(Screen::Credits), show_credits_screen);
-    app.add_systems(OnExit(Screen::Credits), disable_soundtrack);
+    app.add_systems(
+        OnEnter(Screen::Credits),
+        (credits_screen.spawn(), start_credits_soundtrack),
```
janhohenheim, line 15: Calling `spawn()` here and in similar places *looks* like a bug. I know it isn't, but it very very much looks like we accidentally just called the spawn method.

janhohenheim (changes requested):

src/spawn.rs
benfrankel: These both spawn a new entity, though. The difference is whether we're passing in data.

janhohenheim: Oh wait, you're right. Can a system implement `EntityCommand` so that we can remove this function?

src/screens/credits.rs
benfrankel: Well, originally I was going to leave the extension trait that adds this out of the PR, in which case you would have to write a little `fn spawn_credits_screen` system every time. Which isn't too bad. I'm fine with removing `SpawnSystemExt`.

src/spawn.rs
benfrankel: Due to the orphan rule, that would have to be done upstream. But it would be nice if so.

janhohenheim: Fair enough. We still need a better name though.

src/screens/credits.rs
janhohenheim: I think that would help reduce the complexity a bit.

src/screens/playing.rs
benfrankel: Done: https://github.com/TheBevyFlock/bevy_quickstart/pull/231/commits/f891472c0e0eeeeb1ad89d722ca4eb1a7ff76440

src/spawn.rs
janhohenheim: Actually, in this case you *can* bypass the orphan rule by wrapping the system in a struct and then implementing traits on that struct. Not an option for the template because that is severe type magic, but just wanted to let you know.

janhohenheim:
@alice-i-cecile could you also take a look at this?

src/screens/credits.rs
benfrankel: Removed `SpawnSystemExt`.

src/screens/playing.rs
benfrankel: An explanation in the design patterns doc should mitigate that a bit, but I'm open to bikeshedding.

alice-i-cecile (changes requested):
I really don't like this, especially in learning material. There's a *ton* of complexity and type magic here, and it makes simple tasks appear esoteric and difficult. Boilerplate is better than complexity for this.

I understand the desire to improve things here, but I would be strongly opposed to endorsing code with these patterns as official learning materials.

janhohenheim:
@alice-i-cecile I'm tending the same way. I feel a bit bad since we both gave a thumbs-up for it in https://github.com/TheBevyFlock/bevy_quickstart/issues/223, but seeing it in practice is just too "scary". I've been thinking about ways to improve the design and make it more approachable, but I'm really struggling with that.

It would be easier if

- this was an upstreamed API, and so was able to be simplified due to no orphan rule, or
- this was a library, and so a user is opting into it and ready to deal with the new concepts.

For the moment, I'd instead go for the simpler route of replacing observers with custom commands.

benfrankel:
Custom commands means having to extract system parameters from `World`, which is even worse in that regard, no?

janhohenheim:
I don't like extracting stuff manually from `World`, but at least that is "standard Bevy", for better or worse.
Using an `Entity` as the first param is a good analogy to observers, agreed. I personally don't mind it, but I think the Bevy maintainers will disagree with introducing this as an abstraction.

My main issues are more:
- Using the provided abstraction requires active learning, which I don't think is as bad for the other abstractions we use. You can take a single look at `fn plugin` or at `commands.button` and you know how to use it. The same cannot be said for things like the interplay / differences between `spawn_with`, `spawn_fn`, `add_fn`.
- `spawn.rs` feels scary enough that for newcomers it might as well be a black box imo.

As for `Commands` vs `Observer`s: I only recommend using commands because that was @alice-i-cecile's feedback. I personally would prefer an observer because of the ergonomics, even when that means a layer of indirection. My opinion on this is not strong however and I will accept either implementation. I can see both points.

benfrankel:
For the record, custom commands still requires `spawn_with` and extension traits. It would only mean removing `spawn_fn` and `add_fn` and extracting everything from `World` instead of using system parameters.

janhohenheim:
@benfrankel I think that could work, in terms of not introducing too many new abstractions / concepts.

Worst case we can still use the crappy variant that does not return an `Entity` to the caller. It's suboptimal, but it's not our fault that this is fundamentally missing in Bevy and it _might_ be fixed by BSN.

benfrankel:
Done. Usability is much worse without system param support IMO but it's w/e.

src/spawn.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn spawn_with<M: 'static>(&mut self, command: impl EntityCommand<M>) -> EntityWorldMut {
+        let mut e = self.spawn_empty();
+        e.add(command);
+        e
+    }
+}
+
+/// An extension trait that supports arbitrary [`EntityCommand`]s after `world.spawn()`.
+pub trait AddExt {
+    // Workaround for https://github.com/bevyengine/bevy/issues/14278.
```
janhohenheim, line 64:
```suggestion
    // Workaround for <https://github.com/bevyengine/bevy/issues/14278>.
```
Link is currently broken because of the period.

src/spawn.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn spawn_with<M: 'static>(&mut self, command: impl EntityCommand<M>) -> EntityCommands {
+        let mut e = self.spawn_empty();
+        e.add(command);
+        e
+    }
+}
+
+/// An extension trait that provides helper functions for immediate entity spawning.
+pub trait WorldSpawnExt {
+    // Workaround for https://github.com/bevyengine/bevy/issues/14231#issuecomment-2216321086.
```
janhohenheim, line 41:
```suggestion
    // Workaround for <https://github.com/bevyengine/bevy/issues/14231#issuecomment-2216321086>.
```
Link does not point to the comment due to the period

src/spawn.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Re-exported extension traits.
+#[allow(unused_imports)]
+pub mod prelude {
+    pub use super::{AddExt as _, SpawnExt as _, WorldSpawnExt as _};
+}
+
+/// An extension trait that provides helper functions for deferred entity spawning.
+pub trait SpawnExt {
+    // Workaround for https://github.com/bevyengine/bevy/issues/14231#issuecomment-2216321086.
```
janhohenheim, line 18:
```suggestion
    // Workaround for <https://github.com/bevyengine/bevy/issues/14231#issuecomment-2216321086>.
```
Link does not point to the comment due to the period

src/demo/player.rs
```diff
@@ -10,62 +6,57 @@ use crate::{
...
-    image_handles: Res<ImageHandles>,
-    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
-) {
-    // A texture atlas is a way to split one image with a grid into multiple
-    // sprites. By attaching it to a [`SpriteBundle`] and providing an index, we
-    // can specify which section of the image we want to see. We will use this
-    // to animate our player character. You can learn more about texture atlases in
-    // this example: https://github.com/bevyengine/bevy/blob/latest/examples/2d/texture_atlas.rs
+/// Construct a player entity.
+pub fn player(id: Entity, world: &mut World) {
```
janhohenheim, line 27: Could we rename these to `construct_player` and `construct_level` now? Seeing as the underlying pattern has shifted.
`children.spawn_with(construct_player);` sounds completely reasonable to be.

src/demo/player.rs
```diff
@@ -10,62 +6,57 @@ use crate::{
...
-    mut commands: Commands,
-    image_handles: Res<ImageHandles>,
-    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
-) {
-    // A texture atlas is a way to split one image with a grid into multiple
-    // sprites. By attaching it to a [`SpriteBundle`] and providing an index, we
-    // can specify which section of the image we want to see. We will use this
-    // to animate our player character. You can learn more about texture atlases in
-    // this example: https://github.com/bevyengine/bevy/blob/latest/examples/2d/texture_atlas.rs
+/// Construct a player entity.
```
janhohenheim, line 26:
```suggestion
/// Construct a player entity by inserting the relevant components.
```

janhohenheim (changes requested):
I think this version is *way* clearer for an outsider. Other than some minor issues I noted, I'd merge this. 
But I also want to hear @alice-i-cecile's opinion again, as this is controversial enough imo. 
I want to stress that the underlying spawning API is missing from Bevy, so we *will* either have a somewhat complex, but decently usable pattern, or a very simple pattern that will break down when the user experiments with it.
I think this PR strikes a good balance. I'd prefer to ship *some* way of spawning stuff while returning `Entity` or `EntityCommands` or similar as, in contrast to many other tradeoffs of the template, this is *not* something where a user can just switch to a popular community plugin when they are running into trouble.

src/demo/level.rs
```diff
@@ -1,18 +1,23 @@
...
+#[derive(Component, Reflect, Clone, Copy, PartialEq, Eq, Debug)]
+#[reflect(Component)]
+pub struct Level;
 
-fn spawn_level(_trigger: Trigger<SpawnLevel>, mut commands: Commands) {
-    // The only thing we have in our level is a player,
-    // but add things like walls etc. here.
-    commands.trigger(SpawnPlayer);
+/// Construct a level entity.
+pub fn level(id: Entity, world: &mut World) {
```
janhohenheim, line 16: Same as for `fn player`

src/ui/widget.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl Label {
+    pub fn new(text: impl Into<String>) -> Self {
+        Self { text: text.into() }
+    }
+}
+
+/// Construct a UI node that fills the window and centers its children.
+pub fn ui_root(id: Entity, world: &mut World) {
```
janhohenheim, line 211: same as `fn player`

janhohenheim (changes requested):

src/spawn.rs
benfrankel: Where does the link not work? It works for me in VS Code like this.

src/demo/player.rs
benfrankel: I still think that's more confusing (verbosity aside).

src/demo/player.rs
benfrankel: This could add children / do other things as well.

janhohenheim: I would like to guide users more on what "construct" means in this context and why there is a `Entity` as a parameter.

src/spawn.rs
janhohenheim: Clicking it on my Firefox redirects me to this page: https://github.com/bevyengine/bevy/issues/14278.?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen
I think your VS Code is just smart enough to figure out that the dot is not actually part of the URL, while my browser / GitHub assumes it is.

src/demo/player.rs
janhohenheim: I think we are at an impasse with the naming here. Requesting a third opinion from anyone, I'll gladly accept the majority vote.

src/spawn.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Re-exported extension traits.
+#[allow(unused_imports)]
+pub mod prelude {
+    pub use super::{AddExt as _, SpawnExt as _, WorldSpawnExt as _};
+}
+
+/// An extension trait to support spawning an entity from an [`EntityCommand`].
+pub trait SpawnExt {
+    // Workaround for <https://github.com/bevyengine/bevy/issues/14231#issuecomment-2216321086>.
```
alice-i-cecile, line 18: IMO this should just be fixed upstream.

alice-i-cecile:
I think this is an improvement, but I'm still not happy with the level of abstraction / ad-hoc tools here. I think this should be merged, but the linked bug / limitation should be fixed upstream (ping me for reviews!), and I'm still reluctant to officially recommend this template until this is substantially simpler.

If you feel that the limitations of just doing things the dumb way are unacceptable, I'm fine to hold off on endorsing this until the BSN work is shipped.

janhohenheim:
Closing this for now. Sorry again for the whole work that is not merged. Let's revisit this after some improvements have been implemented upstream.

benfrankel:
All good, it was mostly copy/paste and some mechanical refactoring.

Issue #234, opened 2024-08-06
Add OnPress to button widget


Freyja-moth:
Since each usage of the button widget has to add an on press afterwards it might be good to add an OnPress parameter to the method. 

This would also allow to take advantage of impl traits to reduce the footprint of spawning buttons if From<SystemId> is implemented for OnPress.

janhohenheim:
Good idea! We can even make the param a `SystemId` directly, that way `OnPress` can be a hidden implementation detail :)

benfrankel:
It can be a closure that gets registered internally as well.

janhohenheim:
@benfrankel that's fancy 👀

Freyja-moth:
@benfrankel just to check, are you talking about passing in a closure to the buttons method or am I thinking of something completely different?

benfrankel:
Yeah, exactly that.

PR #235, opened 2024-08-06
Fix memory leak using one-shot systems

janhohenheim:
Also implements #234 

Tried adding a closure directly to `fn button`, but that is rather cumbersome, as `ChildBuilder` does not have a `register_one_shot_system` method directly, which means we would have to re-implement it:
```rust
let entity = self.spawn_empty().id();
self.push(RegisterSystem::new(system, entity));
SystemId::from_entity(entity)
```
which is very very *meh*

### DIFF
--- a/src/ui/widgets.rs
+++ b/src/ui/widgets.rs
@@ -1,13 +1,17 @@
 //! Helper traits for creating common widgets.
 
-use bevy::{ecs::system::EntityCommands, prelude::*, ui::Val::*};
+use bevy::{
+    ecs::system::{EntityCommands, SystemId},
+    prelude::*,
+    ui::Val::*,
+};
 
 use super::{interaction::InteractionPalette, palette::*};
 
 /// An extension trait for spawning UI widgets.
 pub trait Widgets {
     /// Spawn a simple button with text.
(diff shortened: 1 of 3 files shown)


src/ui/widgets.rs
```diff
@@ -50,6 +54,9 @@ impl<T: Spawn> Widgets for T {
                 ),
             ));
         });
+        // Add the one-shot system as a child so that will get cleaned up when the button is destroyed.
+        entity.add_child(on_press.entity());
```
janhohenheim, line 58: This here is the line that fixes the leak, as otherwise we never clean up the entities holding our one-shot systems but continue to register new ones when the menu is reconstructed.

src/ui/widgets.rs
```diff
@@ -50,6 +54,9 @@ impl<T: Spawn> Widgets for T {
                 ),
             ));
         });
+        // Add the one-shot system as a child so that will get cleaned up when the button is destroyed.
```
benfrankel, line 57:
```suggestion
        // Add the one-shot system as a child so that it can be despawned when the button despawns.
```

benfrankel (approved):
Wording nit

PR #236, opened 2024-08-06
Reorganize `game` module

janhohenheim:
This makes the `README.md` outdated. I'll fix that in a follow-up.


src/demo/mod.rs
```diff
@@ -0,0 +1,19 @@
+//! Demo gameplay. All of these modules are intented do be here only for demonstration purposes
```
benfrankel, line 1:
```suggestion
//! Demo gameplay. All of these modules are only intended for demonstration purposes
```

src/demo/movement.rs
```diff
@@ -3,6 +3,7 @@
 //!
 //! In our case, the character controller has the following logic:
 //! - Set [`MovementController`] intent based on directional keyboard input.
+//!     This is done in the `player` module, as it is specific to the player character.
```
benfrankel, line 6: Well, unless we add `With<Player>` to the query, _technically_ it's not specific to the player character rn.

src/screens/playing.rs
```diff
@@ -3,9 +3,8 @@
 use bevy::{input::common_conditions::input_just_pressed, prelude::*};
 
 use super::Screen;
-use crate::game::{
-    assets::SoundtrackHandles, audio::soundtrack::SoundtrackCommands as _, spawn::level::SpawnLevel,
-};
+use crate::demo::level::SpawnLevel;
+use crate::{assets::SoundtrackHandles, audio::soundtrack::SoundtrackCommands as _};
```
benfrankel, line 7: I like 1 line per import personally, but should merge these to stay consistent

src/ui/widgets.rs
```diff
@@ -6,7 +6,7 @@ use bevy::{
     ui::Val::*,
 };
 
-use super::{interaction::InteractionPalette, palette::*};
+use super::{interaction::InteractionPalette, palette::*, prelude::OnPress};
```
benfrankel, line 9:
```suggestion
use super::{interaction::{InteractionPalette, OnPress}, palette::*};
```

benfrankel (changes requested):

src/screens/playing.rs
janhohenheim: Ran nightly cargo fmt.

src/ui/widgets.rs
janhohenheim: Ran nightly cargo fmt.

src/demo/movement.rs
janhohenheim: [`ff2dbb3` (#236)](https://github.com/TheBevyFlock/bevy_quickstart/pull/236/commits/ff2dbb31f444d407c9a6f9442b0e91f7adec2f4f)

src/ui/widgets.rs
benfrankel: This isn't resolved by `fmt`, would need to change the import to be from `interaction` instead of `prelude` first.

janhohenheim: Oh, right. Thanks.

benfrankel (approved):

PR #240, opened 2024-08-07
Be more data driven

janhohenheim:
This code should make it more clear that the asset key can be loaded from disk. 
I tried to limit my use of iterators in this PR because they tend to be complicated for beginners, but now there is a bit of a mix between procedural and iterator code. If you want, I can replace it all with procedural code.

Requesting a review from @alice-i-cecile as well.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -25,22 +25,26 @@ pub(super) fn plugin(app: &mut App) {
 pub struct ImageHandles(HashMap<String, Handle<Image>>);
 
 impl ImageHandles {
-    pub const KEY_DUCKY: &'static str = "Ducky";
+    pub const KEY_DUCKY: &'static str = "images/ducky.png";
 }
 
 impl FromWorld for ImageHandles {
     fn from_world(world: &mut World) -> Self {
         let asset_server = world.resource::<AssetServer>();
-        let map = [(
-            ImageHandles::KEY_DUCKY.to_string(),
-            asset_server.load_with_settings(
-                "images/ducky.png",
-                |settings: &mut ImageLoaderSettings| {
-                    settings.sampler = ImageSampler::nearest();
-                },
-            ),
-        )]
-        .into();
+        let pixel_art_settings = |settings: &mut ImageLoaderSettings| {
+            // Using nearest sampling to preserve pixel art style
+            settings.sampler = ImageSampler::nearest();
+        };
+        let pixel_art_files = [ImageHandles::KEY_DUCKY];
+        let map = pixel_art_files
+            .into_iter()
+            .map(|key| {
+                (
+                    key.to_string(),
+                    asset_server.load_with_settings(key, pixel_art_settings),
+                )
+            })
+            .collect();
         Self(map)
     }
 }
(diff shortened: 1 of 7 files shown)


alice-i-cecile (approved):

benfrankel (commented):
If the keys are always paths, I'm wondering if `asset_server.load(ImagePaths::DUCKY)` could replace `image_handles[&ImageHandles::KEY_DUCKY]`? And if not, why not?

I know SFX require a `Vec`, but technically wouldn't it be even more data-driven if it was a single `Handle<Vec<AudioSource>>` somehow? Then the _number of variations_ would also be data-driven.

janhohenheim:
@benfrankel that would work if you wrap `Vec<AudioSource>>` in a type that implements `Asset`. I can try and see how it feels.

janhohenheim:
Update: if I add a custom resource like `SfxSources(Vec<AudioSource>)`, it would require creation by moving all SFX `AudioSource`s out of `Assets` and into `SfxSources`, which only works *after* the SFX sources are done loading, so we need add an event reader for `AssetEvent` and move there.
That, or we create an own `AssetLoader` with its own logic for reading all files, which is not pretty.

Given the amount of complexity this would add, I'm voting for keeping it as-is. We could do `asset_server.load(FOO_KEY)` for non-SFX, but I don't want to introduce inconsistency.

src/assets.rs
```diff
@@ -25,22 +25,23 @@ pub(super) fn plugin(app: &mut App) {
...
-                "images/ducky.png",
-                |settings: &mut ImageLoaderSettings| {
-                    settings.sampler = ImageSampler::nearest();
-                },
-            ),
-        )]
-        .into();
+
+        let pixel_art_settings = |settings: &mut ImageLoaderSettings| {
+            // Using nearest sampling to preserve pixel art style
```
benfrankel, line 36:
```suggestion
            // Use `nearest` image sampling to preserve the pixel art style.
```

src/assets.rs
```diff
@@ -25,22 +25,23 @@ pub(super) fn plugin(app: &mut App) {
...
+            settings.sampler = ImageSampler::nearest();
+        };
+        let mut map = HashMap::new();
+        map.insert(
+            Self::KEY_DUCKY.to_string(),
+            asset_server.load_with_settings(Self::KEY_DUCKY, pixel_art_settings),
+        );
```
benfrankel, line 43: This could be the same `[key].into_iter().map().collect()` as the other two structs.

benfrankel (changes requested):

src/assets.rs
benfrankel: Or we could mention `ImageFilterMode::Nearest`? That's probably overkill though.

src/assets.rs
janhohenheim: [`f1b7382` (#240)](https://github.com/TheBevyFlock/bevy_quickstart/pull/240/commits/f1b73827958847f8b3759ecccefa02296362b789)

benfrankel (approved):

PR #242, opened 2024-08-07
Use custom commands instead of observers for spawning

janhohenheim:
Part of #203
Alternative to #231
Does *not* help with #223.

### DIFF
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -1,8 +1,8 @@
 //! Spawn the main level by triggering other observers.
 
-use bevy::prelude::*;
+use bevy::{ecs::world::Command, prelude::*};
 
-use super::player::SpawnPlayerCommands as _;
+use super::player::SpawnPlayer;
 
 pub(super) fn plugin(_app: &mut App) {
     // No setup required for this plugin.
@@ -10,14 +10,13 @@ pub(super) fn plugin(_app: &mut App) {
     // later if needed.
 }
 
-pub trait SpawnLevelCommands {
-    fn spawn_level(&mut self);
-}
+#[derive(Debug)]
+pub struct SpawnLevel;
 
-impl SpawnLevelCommands for Commands<'_, '_> {
-    fn spawn_level(&mut self) {
+impl Command for SpawnLevel {
+    fn apply(self, world: &mut World) {
         // The only thing we have in our level is a player,
         // but add things like walls etc. here.
-        self.spawn_player();
+        world.commands().add(SpawnPlayer);
     }
 }
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -17,6 +17,6 @@ impl Command for SpawnLevel {
     fn apply(self, world: &mut World) {
         // The only thing we have in our level is a player,
         // but add things like walls etc. here.
-        world.commands().add(SpawnPlayer::default());
+        world.commands().add(SpawnPlayer { max_speed: 400.0 });
     }
 }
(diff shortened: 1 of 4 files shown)


janhohenheim:
@alice-i-cecile is this what you had in mind for the simple but limited spawning? Or would you prefer if I used a struct `SpawnPlayer` that the user then manually `add`s?

src/demo/level.rs
```diff
@@ -1,18 +1,22 @@
 //! Spawn the main level by triggering other observers.
```
benfrankel, line 1: Outdated. Also, implementation details shouldn't be exposed in module-level docs.

src/demo/player.rs
```diff
@@ -2,7 +2,10 @@
 //! Note that this is separate from the `movement` module as that could be used
 //! for other characters as well.
 
-use bevy::prelude::*;
+use bevy::{
+    ecs::{system::RunSystemOnce, world::Command},
```
benfrankel, line 6:
```suggestion
    ecs::{system::RunSystemOnce as _, world::Command},
```
Optional.

src/demo/player.rs
```diff
@@ -25,15 +27,22 @@ pub(super) fn plugin(app: &mut App) {
...
 
-#[derive(Event, Debug)]
-pub struct SpawnPlayer;
-
 #[derive(Component, Debug, Clone, Copy, PartialEq, Eq, Default, Reflect)]
 #[reflect(Component)]
 pub struct Player;
 
+/// Command to spawn the player character.
+/// We can add fields to this struct if we need data for spawning.
```
benfrankel, line 35:
```suggestion
/// A command to spawn the player character.
```
I don't like the second line here personally. There are a lot of things the user _could_ do to expand on the template, and we're not going to list them all. At the very least this doesn't make sense as a doc comment.

src/demo/level.rs
janhohenheim: [`e56ec7d` (#242)](https://github.com/TheBevyFlock/bevy_quickstart/pull/242/commits/e56ec7d77cf71e25957602d5abe7b94b441c1895)

src/demo/player.rs
janhohenheim: We do however mention things that have a very high probability of coming up and are low-hanging expansions.
Moved the comment into the struct: [`d2c59b6` (#242)](https://github.com/TheBevyFlock/bevy_quickstart/pull/242/commits/d2c59b6ef9162e66971ade4188e99d4246ee75b0)

benfrankel (approved):
Approved. I think `spawn_level` should spawn a parent level entity with a `Level` marker component ideally, but this PR doesn't have to change the current behavior.

src/demo/level.rs
```diff
@@ -1,18 +1,22 @@
...
 
-fn spawn_level(_trigger: Trigger<SpawnLevel>, mut commands: Commands) {
-    // The only thing we have in our level is a player,
-    // but add things like walls etc. here.
-    commands.trigger(SpawnPlayer);
+impl Command for SpawnLevel {
+    fn apply(self, world: &mut World) {
+        // The only thing we have in our level is a player,
+        // but add things like walls etc. here.
+        world.commands().add(SpawnPlayer { max_speed: 400.0 });
```
benfrankel, line 20: Might need to do `world.flush_commands()` after this

benfrankel (approved):

src/demo/level.rs
janhohenheim: [`41c6194` (#242)](https://github.com/TheBevyFlock/bevy_quickstart/pull/242/commits/41c6194bb969721cf5c4bd472c7e482c0fa7132c)

Issue #202, opened 2024-07-30
Add VS Code snippets


benfrankel:
For example, `component` can expand into:

```rust
#[derive(Component, Reflect)]
#[reflect(Component)]
struct _;
```

With cursor positioned on the `_`.

janhohenheim:
Good idea. I'd personally add `Debug` and `reflect(Debug)` since you always want these in practice.

benfrankel:
There are some existing Bevy code snippets in the VS Code Marketplace, but they're limited / archived on GitHub as far as I can tell.

PR #247, opened 2024-08-09
Add VS Code snippets for Bevy

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/202.

### DIFF
--- a/.vscode/bevy.code-snippets
+++ b/.vscode/bevy.code-snippets
@@ -1,35 +1,32 @@
 {
-    "Bevy: New module-level function Plugin": {
+    "Bevy: New top-level function Plugin": {
         "scope": "rust",
-        "prefix": "plugin_fn",
+        "prefix": "plugin",
         "body": [
             "use bevy::prelude::*;",
             "",
             "pub(super) fn plugin(app: &mut App) {",
             "\t$0",
             "}"
-        ],
-        "description": "Define a new module-level function Plugin"
+        ]
     },
     "Bevy: New Component": {
         "scope": "rust",
         "prefix": "component",
         "body": [
             "#[derive(Component, Reflect, Debug)]",
-            "#[reflect(Component, Debug)]",
+            "#[reflect(Component)]",
             "struct $1;"
-        ],
-        "description": "Define a new Component"
+        ]
     },
     "Bevy: New Resource": {
         "scope": "rust",
         "prefix": "resource",
         "body": [
             "#[derive(Resource, Reflect, Debug, Default)]",
-            "#[reflect(Resource, Debug)]",
+            "#[reflect(Resource)]",
             "struct $1;"
-        ],
-        "description": "Define a new Resource"
+        ]
     },
     "Bevy: New Event": {
         "scope": "rust",
@@ -37,19 +34,17 @@
         "body": [
             "#[derive(Event, Debug)]",
             "struct $1;"
-        ],
-        "description": "Define a new Event"
+        ]
     },
     "Bevy: New SystemSet": {
         "scope": "rust",
-        "prefix": "system_set",
+        "prefix": "systemset",
         "body": [
             "#[derive(SystemSet, Copy, Clone, Eq, PartialEq, Hash, Debug)]",
             "enum $1 {",
             "\t$0",
             "}"
-        ],
-        "description": "Define a new SystemSet"
+        ]
     },
     "Bevy: New Schedule": {
         "scope": "rust",
@@ -57,8 +52,7 @@
         "body": [
             "#[derive(ScheduleLabel, Copy, Clone, Eq, PartialEq, Hash, Debug)]",
             "struct $1;"
-        ],
-        "description": "Define a new Schedule"
+        ]
     },
     "Bevy: New States": {
         "scope": "rust",
@@ -69,7 +63,6 @@
             "\t#[default]",
             "\t$0",
             "}"
-        ],
-        "description": "Define a new States"
+        ]
     }
 }
--- a/.vscode/bevy.code-snippets
+++ b/.vscode/bevy.code-snippets
@@ -1,0 +1,75 @@
+{
+    "Bevy: New module-level function Plugin": {
+        "scope": "rust",
+        "prefix": "plugin_fn",
+        "body": [
+            "use bevy::prelude::*;",
+            "",
+            "pub(super) fn plugin(app: &mut App) {",
+            "\t$0",
+            "}"
+        ],
+        "description": "Define a new module-level function Plugin"
+    },
+    "Bevy: New Component": {
+        "scope": "rust",
+        "prefix": "component",
+        "body": [
+            "#[derive(Component, Reflect)]",
+            "#[reflect(Component)]",
+            "struct $1;"
+        ],
+        "description": "Define a new Component"
+    },
+    "Bevy: New Resource": {
+        "scope": "rust",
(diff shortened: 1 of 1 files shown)


.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
+{
+    "Bevy: New module-level function Plugin": {
+        "scope": "rust",
+        "prefix": "plugin_fn",
```
janhohenheim, line 4:
```suggestion
        "prefix": "plugin",
```
works well enough in my experience.

.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
...
+        "body": [
+            "use bevy::prelude::*;",
+            "",
+            "pub(super) fn plugin(app: &mut App) {",
+            "\t$0",
```
janhohenheim, line 9: what does `$0` do for a prefix?

.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
...
+        "scope": "rust",
+        "prefix": "states",
+        "body": [
+            "#[derive(States, Copy, Clone, Default, Eq, PartialEq, Hash, Debug, Reflect)]",
+            "enum $1 {",
+            "\t#[default]",
+            "\t$0",
+            "}"
+        ],
+        "description": "Define a new States"
```
janhohenheim, line 73:
```suggestion
        "description": "Define new States"
```
I know why you wrote it the original way, but it reads weird

.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
...
+            "\t$0",
+            "}"
+        ],
+        "description": "Define a new module-level function Plugin"
+    },
+    "Bevy: New Component": {
+        "scope": "rust",
+        "prefix": "component",
+        "body": [
+            "#[derive(Component, Reflect)]",
```
janhohenheim:
```suggestion
            "#[derive(Component, Reflect, Debug)]",
```

.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
...
+            "#[reflect(Component)]",
+            "struct $1;"
+        ],
+        "description": "Define a new Component"
+    },
+    "Bevy: New Resource": {
+        "scope": "rust",
+        "prefix": "resource",
+        "body": [
+            "#[derive(Resource, Reflect, Default)]",
```
janhohenheim:
```suggestion
            "#[derive(Resource, Reflect, Default, Debug)]",
```

.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
...
+            "#[reflect(Resource)]",
+            "struct $1;"
+        ],
+        "description": "Define a new Resource"
+    },
+    "Bevy: New Event": {
+        "scope": "rust",
+        "prefix": "event",
+        "body": [
+            "#[derive(Event)]",
```
janhohenheim:
```suggestion
            "#[derive(Event, Debug)]",
```

.vscode/bevy.code-snippets
```diff
@@ -0,0 +1,75 @@
+{
+    "Bevy: New module-level function Plugin": {
+        "scope": "rust",
+        "prefix": "plugin_fn",
+        "body": [
+            "use bevy::prelude::*;",
```
janhohenheim, line 6: Remove this line and use `"requires": "bevy::prelude::*"` instead. Add that to all other snippets as well.

janhohenheim (changes requested):
Many of these could be `postfix` instead, but I have no strong feelings on that.
Does adding these preserve the original custom Rust snippets? I know that they can get overridden pretty easily if you're not careful. For reference, these are the builtin ones:
```json
       "Arc::new": {
            "postfix": "arc",
            "body": "Arc::new(${receiver})",
            "requires": "std::sync::Arc",
            "description": "Put the expression into an `Arc`",
            "scope": "expr"
        },
        "Rc::new": {
            "postfix": "rc",
            "body": "Rc::new(${receiver})",
            "requires": "std::rc::Rc",
            "description": "Put the expression into an `Rc`",
            "scope": "expr"
        },
        "Box::pin": {
            "postfix": "pinbox",
            "body": "Box::pin(${receiver})",
            "requires": "std::boxed::Box",
            "description": "Put the expression into a pinned `Box`",
            "scope": "expr"
        },
        "Ok": {
            "postfix": "ok",
            "body": "Ok(${receiver})",
            "description": "Wrap the expression in a `Result::Ok`",
            "scope": "expr"
        },
        "Err": {
            "postfix": "err",
            "body": "Err(${receiver})",
            "description": "Wrap the expression in a `Result::Err`",
            "scope": "expr"
        },
        "Some": {
            "postfix": "some",
            "body": "Some(${receiver})",
            "description": "Wrap the expression in an `Option::Some`",
            "scope": "expr"
        }
 ```

.vscode/bevy.code-snippets
benfrankel: We might want a struct Plugin snippet at some point?

.vscode/bevy.code-snippets
benfrankel: That's where the cursor ends up after you Tab through all the placeholder locations (`$1, $2, ...`).

.vscode/bevy.code-snippets
benfrankel: It reads weird because `bevy_state` is weird ^\\\_(\`\`/)\_/^

.vscode/bevy.code-snippets
benfrankel: Woah

.vscode/bevy.code-snippets
janhohenheim: TIL

.vscode/bevy.code-snippets
janhohenheim: Too niche imo

.vscode/bevy.code-snippets
benfrankel: I get a "Property requires is not allowed" warning, and I don't see this mentioned in the docs.

janhohenheim: Got it from the [rust-analyzer docs](https://rust-analyzer.github.io/manual.html#user-snippet-completions)

.vscode/bevy.code-snippets
janhohenheim: either change it to "Define new States" or "Define a new States enum", either are fine for me.

.vscode/bevy.code-snippets
janhohenheim: Follow-up: I'd prefer to have the more common use-case easy to use (`plugin`) and have the more specialized one be more verbose (`plugin_struct`)

.vscode/bevy.code-snippets
janhohenheim: Update from Discord: the rust-analyzer snippet syntax does *not* work.

.vscode/bevy.code-snippets
benfrankel: I'm going to remove all of the descriptions because they don't provide any value over the names. It looks like this with no description:
2024-08-09_1723241293_863x147

janhohenheim (approved):

Issue #201, opened 2024-07-29
Workspace recommended extensions


wduminy:
Would be nice to have the extensions mentioned in [tooling.md](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions) specified as [workspace recommended extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions).

janhohenheim:
TIL about that feature, thanks!

PR #248, opened 2024-08-09
Add workspace recommended extensions for VS Code

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/201.


janhohenheim (approved):
Haven't tested if this actually works, but looks reasonable :)

PR #249, opened 2024-08-09
Use observers for callbacks

janhohenheim:
- This variant is probably closer to what will be upstreamed in Bevy, as Alice mentioned wanting to solve this using observers.
- Its implementation is simpler (no need to learn about one-shot systems or `SystemId`, no need to be careful about memory leaks)
- I removed the extra param from `button`. 
  - With the current code, the user does not save on boilerplate when passing the callback to the function. But it incurs a bit of mental indirection that is not present when `observe` is called directly and it requires us to add some ugly generics to the `button` function. I like about the current code that it kinda teaches the user how to apply observers to entities for themselves.
  - Of course, this means a user can forget to handle the `OnPress` event as a consequence.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -254,6 +254,20 @@ The event `OnPress`, which is [defined in this template](../src/theme/interactio
 is triggered when the button's [`Interaction`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html)
 component becomes [`Interaction::Pressed`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html#variant.Pressed).
 
+If your interactions mostly consist of changing the game state, consider using the following helper function:
+
+```rust
+fn spawn_button(mut commands: Commands) {
+    commands.button("Play the game").observe(change_state(Screen::Playing));
+}
+
+fn change_state<S: FreelyMutableState>(
+    new_state: S,
+) -> impl Fn(Trigger<OnPress>, ResMut<NextState<S>>) {
+    move |_trigger, mut next_state| next_state.set(new_state.clone())
+}
+```
+
 ### Reasoning
 
 This pattern is inspired by [bevy_mod_picking](https://github.com/aevyrie/bevy_mod_picking).
--- a/docs/design.md
+++ b/docs/design.md
@@ -270,7 +270,7 @@ fn change_state<S: FreelyMutableState>(
 ### Reasoning
 
 This pattern is inspired by [bevy_mod_picking](https://github.com/aevyrie/bevy_mod_picking).
-By coupling the system handling the interaction to the entity as an observer,
+By pairing the system handling the interaction with the entity as an observer,
 the code running on interactions can be scoped to the exact context of the interaction.
 
 For example, the code for what happens when you press a *specific* button is directly attached to that exact button.
--- a/docs/design.md
+++ b/docs/design.md
@@ -276,8 +276,7 @@ the code running on interactions can be scoped to the exact context of the inter
 For example, the code for what happens when you press a *specific* button is directly attached to that exact button.
 
 This also keeps the interaction logic close to the entity that is interacted with,
-allowing for better code organization. If you want multiple buttons to do the same thing,
-consider triggering a secondary event in their callbacks.
+allowing for better code organization.
 
 ## Dev Tools
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -251,8 +251,7 @@ fn pay_money(_trigger: Trigger<OnPress>, /* Bevy query parameters */) {
 ```
 
 The event `OnPress`, which is [defined in this template](../src/theme/interaction.rs),
-is triggered when the button's [`Interaction`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html)
-component becomes [`Interaction::Pressed`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html#variant.Pressed).
+is triggered when the button is [`Interaction::Pressed`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html#variant.Pressed).
 
 If your interactions mostly consist of changing the game state, consider using the following helper function:
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -245,8 +245,8 @@ fn spawn_button(mut commands: Commands) {
     commands.button("Pay up!").observe(pay_money);
 }
 
-fn pay_money(_trigger: Trigger<OnPress>, /* Bevy query parameters */) {
-    // Handle the interaction
+fn pay_money(_trigger: Trigger<OnPress>, mut money: ResMut<Money>) {
+    money.0 -= 10.0;
 }
 ```
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -257,10 +257,10 @@ If you have many interactions that only change a state, consider using the follo
 
 ```rust
 fn spawn_button(mut commands: Commands) {
-    commands.button("Play the game").observe(change_state(Screen::Playing));
+    commands.button("Play the game").observe(enter_state(Screen::Playing));
 }
 
-fn change_state<S: FreelyMutableState>(
+fn enter_state<S: FreelyMutableState>(
     new_state: S,
 ) -> impl Fn(Trigger<OnPress>, ResMut<NextState<S>>) {
     move |_trigger, mut next_state| next_state.set(new_state.clone())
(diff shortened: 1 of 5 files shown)


src/screens/title.rs
```diff
@@ -10,32 +10,27 @@ pub(super) fn plugin(app: &mut App) {
...
-fn enter_credits(mut next_screen: ResMut<NextState<Screen>>) {
+fn enter_credits(_trigger: Trigger<OnPress>, mut next_screen: ResMut<NextState<Screen>>) {
     next_screen.set(Screen::Credits);
 }
 
 #[cfg(not(target_family = "wasm"))]
-fn exit_app(mut app_exit: EventWriter<AppExit>) {
+fn exit_app(_trigger: Trigger<OnPress>, mut app_exit: EventWriter<AppExit>) {
     app_exit.send(AppExit::Success);
 }
```
MiniaczQ, line 36: Could we use closures for this?

MiniaczQ: A button setting a state seems like a pretty generic use case

janhohenheim: I would not do that in this case, as I want to teach the user how to do stuff on button presses in general, and a closure wrapper would obscure that somewhat.

MiniaczQ: Can we replace just one of those, to show the generic change state or at least mention it in a comment how it scales?
Users might try to re-use the same functions between different themes, which is much worse than a single generic system
And you can have multiple observers on a button press, so adding more logic doesn't invalidate this case

MiniaczQ: I'll accept "it doesn't need to scale" since most games have very simple navigation, just seeing if you would consider it

benfrankel: pyri_state provides `Screen::Credits.enter()` for this btw. seems like something bevy_state should just provide instead.

MiniaczQ: basically the same thing 🤔 
I feel like functions returning closures is already a pattern so I'd personally stick with that

As for `bevy_state` relying on observers, good idea, I'd just wait for observers to stabilize a bit more

docs/design.md
```diff
@@ -237,28 +237,34 @@ as custom commands don't return `Entity` or `EntityCommands`. This kind of usage
...
 
 ```rust
 fn spawn_button(mut commands: Commands) {
-    let pay_money = commands.register_one_shot_system(pay_money);
-    commands.button("Pay up!", pay_money);
+    // See the Widgets pattern for information on the `button` method
+    commands.button("Pay up!").observe(pay_money);
+}
+
+fn pay_money(_trigger: Trigger<OnPress>, /* Bevy query parameters */) {
```
benfrankel, line 248:
```suggestion
fn pay_money(_trigger: Trigger<OnPress>, /* Bevy system parameters */) {
```
Although this would be more clear with e.g. a dummy `mut money: ResMut<Money>` and then `money.0 += 100.0;` in the body.

docs/design.md
```diff
@@ -237,28 +237,34 @@ as custom commands don't return `Entity` or `EntityCommands`. This kind of usage
...
+    // Handle the interaction
 }
 ```
 
-The resulting `SystemId` is added as a newtype component on the button entity.
-See the definition of [`OnPress`](../src/theme/interaction.rs) for how this is done.
+The event `OnPress`, which is [defined in this template](../src/theme/interaction.rs),
+is triggered when the button's [`Interaction`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html)
+component becomes [`Interaction::Pressed`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html#variant.Pressed).
```
benfrankel, line 255:
```suggestion
is triggered when the button is [`Interaction::Pressed`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html#variant.Pressed).
```
Optional.

docs/design.md
```diff
@@ -237,28 +237,34 @@ as custom commands don't return `Entity` or `EntityCommands`. This kind of usage
...
 
 ### Reasoning
 
 This pattern is inspired by [bevy_mod_picking](https://github.com/aevyrie/bevy_mod_picking).
-By adding the system handling the interaction to the entity as a component,
+By coupling the system handling the interaction to the entity as an observer,
```
benfrankel, line 274:
```suggestion
By pairing the system handling the interaction to the entity as an observer,
```
Optional. "coupling" reminds me of "code coupling" which is an anti-pattern.

src/screens/title.rs
janhohenheim: [`4529b0f` (#249)](https://github.com/TheBevyFlock/bevy_quickstart/pull/249/commits/4529b0fb7a9cd9fbd0a372f85952560ebd022d37)

docs/design.md
```diff
@@ -237,28 +237,34 @@ as custom commands don't return `Entity` or `EntityCommands`. This kind of usage
...
 the code running on interactions can be scoped to the exact context of the interaction.
 
 For example, the code for what happens when you press a *specific* button is directly attached to that exact button.
 
 This also keeps the interaction logic close to the entity that is interacted with,
-allowing for better code organization. If you want multiple buttons to do the same thing, consider triggering an event in their callbacks.
+allowing for better code organization. If you want multiple buttons to do the same thing,
+consider triggering a secondary event in their callbacks.
```
benfrankel, line 281: You can just use the same observer, no? `button1.observe(foo); button2.observe(foo);` where `foo` takes `Trigger<OnPress>`.

benfrankel (changes requested):

MiniaczQ (approved):
This is way better than what was here previously!

docs/design.md
MiniaczQ: I had the same thought, decided to not comment as only "consideration" is suggested 😅

docs/design.md
janhohenheim: Wouldn't it be "pairing A with B" instead of "pairing A to B"?

docs/design.md
```diff
@@ -237,28 +237,47 @@ as custom commands don't return `Entity` or `EntityCommands`. This kind of usage
...
+    // Handle the interaction
 }
 ```
 
-The resulting `SystemId` is added as a newtype component on the button entity.
-See the definition of [`OnPress`](../src/theme/interaction.rs) for how this is done.
+The event `OnPress`, which is [defined in this template](../src/theme/interaction.rs),
+is triggered when the button is [`Interaction::Pressed`](https://docs.rs/bevy/latest/bevy/prelude/enum.Interaction.html#variant.Pressed).
+
+If your interactions mostly consist of changing the game state, consider using the following helper function:
```
benfrankel, line 256:
```suggestion
If you have many interactions that only change a state, consider using the following helper function:
```
Reasoning: "game state" is redundant, every state is a game state; and the interactions have to be _only_ changing the state, not "mostly consisting of" changing the state, for the helper function to be useful.

docs/design.md
```diff
@@ -237,28 +237,47 @@ as custom commands don't return `Entity` or `EntityCommands`. This kind of usage
...
+
+```rust
+fn spawn_button(mut commands: Commands) {
+    commands.button("Play the game").observe(change_state(Screen::Playing));
+}
+
+fn change_state<S: FreelyMutableState>(
```
benfrankel, line 263: Nit: `enter_state` reads better IMO.

docs/design.md
janhohenheim: [`6fd702b` (#249)](https://github.com/TheBevyFlock/bevy_quickstart/pull/249/commits/6fd702b3fda95461e7653a5997d1ea6634f660ab)

docs/design.md
janhohenheim: [`0f27c74` (#249)](https://github.com/TheBevyFlock/bevy_quickstart/pull/249/commits/0f27c74a8c4d91d7c7c469195a32453b6e449ee2)

docs/design.md
janhohenheim: [`6d794b2` (#249)](https://github.com/TheBevyFlock/bevy_quickstart/pull/249/commits/6d794b242347929155499c766cf7846d1aceef8e)

benfrankel (changes requested):

docs/design.md
benfrankel: I think either way is correct.

benfrankel (approved):

docs/design.md
janhohenheim: [`724986e` (#249)](https://github.com/TheBevyFlock/bevy_quickstart/pull/249/commits/724986e4c60592832a20006861ab29b5440193dc)

Issue #200, opened 2024-07-29
Expand the workflows docs to mention the specific permissions needed


TrialDragon:
> It would be nice to document which permission to add, and not to recommend adding a blanket write permission for everything. Anyone has any idea which permission exactly is missing in this particular case?

from djeedai on [Discord](https://discord.com/channels/691052431525675048/937158127491633152/1267169649896980521).

I figured it'd be a good thing to mention here for consideration.

janhohenheim:
I think workflow files may be able to specify their permissions directly, in which case users do not need to tweak the settings at all. Not sure though.

benfrankel:
Yeah I skimmed the documentation and it seems to suggest that. If true, that would be a UX improvement as well.

janhohenheim:
Here's an example in the wild, courtesy of @bash: https://github.com/bash/bash.github.io/blob/main/.github/workflows/ci.yml#L43-L45

PR #252, opened 2024-08-10
Simplify release workflow UX

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/200.


docs/workflows.md
```diff
@@ -46,66 +46,51 @@ Enter a version number in the format `v1.2.3`, then hit the green `Run workflow`
...
+    ASSETS_DIR: assets
+    # Whether packages produced by this workflow should be uploaded to the Github release.
+    UPLOAD_PACKAGES_TO_GITHUB_RELEASE: true
+    # Before enabling LFS, please take a look at GitHub's documentation for costs and quota limits:
+    # https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage
+    USE_GIT_LFS: false
+  ```
 </details>
 
-In any case, make note of the value used for `ITCH_TARGET`, as you will need it later.
```
janhohenheim, line 73: I'd leave this.

docs/workflows.md
```diff
@@ -46,66 +46,51 @@ Enter a version number in the format `v1.2.3`, then hit the green `Run workflow`
...
+</details>
 
-Hit `New repository secret` and enter the following, then hit `Add secret`:
+Hit `New repository secret` and enter the following values, then hit `Add secret`:
 
 - **Name:** `BUTLER_CREDENTIALS`
 - **Secret:** Your [itch.io API key](https://itch.io/user/settings/api-keys) (create a new one if necessary)
 
-#### Select the web build
+#### Set up itch.io project
```
janhohenheim, line 87: This was above so that it was closer to the section that references the `ITCH_TARGET`

janhohenheim (changes requested):

docs/workflows.md
janhohenheim: Hmm, I see how it's simpler this way. Let's leave it like this.

benfrankel: We link to it so no need. Putting this below means that the GitHub repo is fully set up by the time you start setting up your itch.io prjoect, which allows the two sections to be combined into one (Create page + Select the web build), avoiding the awkward "there is an error but we'll fix it later".

docs/workflows.md
benfrankel: I could leave it in, but I'd be surprised if a user got tripped up due to the absence of this line since it's already explained at the step where it's actually used.

janhohenheim: aight, I'll trust you

janhohenheim (approved):

PR #256, opened 2024-08-10
Refactor `env` variables in release workflow

benfrankel:


janhohenheim:
Could you test this with a non `.` workspace before merging?

.github/workflows/release.yaml
```diff
@@ -17,21 +17,21 @@ on:
...
   # The base filename of the binary produced by `cargo build`.
-  BINARY: bevy_quickstart
-  # The name to use for the packaged application produced by this workflow.
-  PACKAGE_NAME: bevy_quickstart
-  # The itch.io page to upload to, in the format: `user-name/project-name`.
-  # Comment this out to disable.
-  ITCH_TARGET: the-bevy-flock/bevy-quickstart
-  # The organization or author that owns the rights to the game.
-  OWNER: the-bevy-flock
+  cargo_binary_name: bevy_quickstart
```
janhohenheim, line 20: This sounds a bit like the name of the `cargo` binary itself now. Maybe `crate_binary_name` would be more accurate?

.github/workflows/release.yaml
```diff
@@ -17,21 +17,21 @@ on:
...
   # The path to the assets directory.
-  ASSETS_DIR: assets
-  # Whether packages produced by this workflow should be uploaded to the Github release.
-  UPLOAD_PACKAGES_TO_GITHUB_RELEASE: true
+  assets: assets
```
janhohenheim, line 22: `assets_dir` sounds less confusing to me.

.github/workflows/release.yaml
```diff
@@ -17,21 +17,21 @@ on:
...
+  # The name to use for the app produced by this workflow.
+  app_name: bevy_quickstart
```
janhohenheim, line 24:
```suggestion
  app_name: Bevy Quickstart
```
This should make it more clear what the difference between this and the other binary name is.

.github/workflows/release.yaml
```diff
@@ -17,21 +17,21 @@ on:
...
+  # The organization or author that owns the rights to the app.
```
janhohenheim, line 25:
```suggestion
  # The organization or author that owns the rights to the app. Must be either kebab-case or snake_case.
```
It's an Apple thing.

janhohenheim (changes requested):

.github/workflows/release.yaml
benfrankel: Nah you want this to be `bevy_quickstart` so that the package is `bevy_quickstart-windows.zip` with a `bevy_quickstart.exe` binary inside, rather than `Bevy Quickstart-windows.zip` with a `Bevy Quickstart.exe` binary inside.

benfrankel: Unless we want to have 2 env variables, one for the MacOS "app name" in the metadata file, and one for the app package / binary filenames.

.github/workflows/release.yaml
benfrankel: Maybe `assets_path`? Since it's the path to the directory, not just the directory name.

.github/workflows/release.yaml
benfrankel: Would `cargo_build_binary_name` be okay? I know it's verbose, but it's only in one place in the release workflow to move the binary produced by `cargo build` out of `target/...`. And it's only needed because it's currently not possible to tell `cargo` where to put the binary: https://github.com/rust-lang/cargo/issues/6790

.github/workflows/release.yaml
benfrankel: Maybe we _do_ need to split `app_name` into two variables, with one of them being specifically for MacOS.

.github/workflows/release.yaml
janhohenheim: that's fine, yeah

.github/workflows/release.yaml
janhohenheim: yeah, that's better

.github/workflows/release.yaml
janhohenheim: At least on Windows, having executables like "Some Cool Executable.exe" with spaces in-between is not that uncommon.
In any case, it currently is a bit confusing that there are two very similar names at play here. One of them could just be defaulted.
The macOS app name should have an own env var if the others are not supposed to have spaces.

.github/workflows/release.yaml
```diff
@@ -127,47 +127,47 @@ jobs:
...
           fi
 
-      - name: Add assets to package (non-Web)
+      - name: Add assets to app (non-Web)
         if: ${{ matrix.platform != 'web' }}
-        run: cp -r '${{ env.ASSETS_DIR }}' '${{ env.OUT_DIR }}' || true # Ignore error if assets folder does not exist
+        run: cp -r '${{ env.assets }}' '${{ env.app }}' || true # Ignore error if assets folder does not exist
 
-      - name: Add app metadata to package (MacOS)
+      - name: Add metadata to app (MacOS)
```
janhohenheim, line 153:
```suggestion
      - name: Add metadata to app (macOS)
```
while we are on it.

The capitalized M is only used in legacy contexts (like the `Contents/MacOS` dir) because it was used for Mac OSX, while the current OS is called macOS

.github/workflows/release.yaml
```diff
@@ -92,8 +92,8 @@ jobs:
     steps:
       - name: Set up environment
         run: |
-          echo 'PACKAGE=${{ env.PACKAGE_NAME }}-${{ matrix.platform }}' >> "${GITHUB_ENV}"
-          echo 'OUT_DIR=tmp/package/${{ env.PACKAGE_NAME }}${{ matrix.out_dir_suffix }}' >> "${GITHUB_ENV}"
+          echo 'app=tmp/app/${{ env.app_name }}${{ matrix.app_suffix }}' >> "${GITHUB_ENV}"
+          echo 'package=${{ env.app_name }}-${{ matrix.platform }}${{ matrix.package_ext }}' >> "${GITHUB_ENV}"
           if [ '${{ matrix.platform }}' == 'macos' ]; then
             echo 'MACOSX_DEPLOYMENT_TARGET=11.0' >> "${GITHUB_ENV}" # MacOS 11.0 Big Sur is the first version to support universal binaries.
```
janhohenheim, line 98:
```suggestion
            echo 'MACOSX_DEPLOYMENT_TARGET=11.0' >> "${GITHUB_ENV}" # macOS 11.0 Big Sur is the first version to support universal binaries.
```

janhohenheim (changes requested):

.github/workflows/release.yaml
benfrankel: It's _allowed_ but very strange to have spaces in a linux executable.

docs/workflows.md
```diff
@@ -47,23 +47,51 @@ The release workflow can be configured by tweaking the environment variables in
...
+
+  # The path to the assets directory.
+  assets_path: assets
+
+  # Whether to upload the packages produced by this workflow to a GitHub release.
+  upload_to_github: true
+
+  # The itch.io project to upload to in the format `user-name/project-name`.
+  # There will be no upload to itch.io if this is commented out.
+  upload_to_itch: the-bevy-flock/bevy-quickstart
```
janhohenheim, line 61: This is still using the old spelling further down this document

janhohenheim (approved):
Minor issue. Feel free to merge when it's fixed :)

docs/workflows.md
benfrankel: Fixed, thank you

PR #257, opened 2024-08-10
Splash screen allows exiting early (with escape key)

will-hart:
This PR allows the player to exit the splash screen at any time by hitting `KeyCode::escape`.

---

Previously this PR also contained code for displaying multiple splash screens, but after discussion that will be migrated to an example in the bevy repository.

### DIFF
--- a/src/screens/splash.rs
+++ b/src/screens/splash.rs
@@ -40,7 +40,7 @@ pub(super) fn plugin(app: &mut App) {
             .run_if(in_state(Screen::Splash)),
     );
 
-    // exit the splash screen early if the player hits escape
+    // Exit the splash screen early if the player hits escape.
     app.add_systems(
         Update,
         exit_splash_screen
@@ -64,7 +64,7 @@ impl FromWorld for SplashScreenImageList {
     fn from_world(world: &mut World) -> Self {
         let asset_server = world.resource::<AssetServer>();
 
-        // To show your own splash images, replace or add images here
+        // To show your own splash images, replace or add images here.
         Self(VecDeque::from_iter(
             [asset_server.load("images/splash.png")],
         ))
@@ -104,10 +104,10 @@ fn spawn_splash(
     mut next_screen: ResMut<NextState<Screen>>,
 ) {
     if splash_images.0.is_empty() {
-        // if there are no splash images, go directly to the loading screen
+        // If there are no splash images, go directly to the loading screen.
         next_screen.set(Screen::Loading);
     } else {
-        // spawn the UI root but wait for the first timer tick to show an image
+        // Spawn the UI root but wait for the first timer tick to show an image.
         commands.ui_root().insert((
             Name::new("Splash screen container"),
             BackgroundColor(SPLASH_BACKGROUND_COLOR),
@@ -184,9 +184,9 @@ fn check_splash_timer(
         return;
     }
 
-    // try to get the next splash image. If there isn't one, we move on to the next screen
+    // Try to get the next splash image. If there isn't one, we move on to the next screen.
     let Some(next_splash_image) = splash_images.0.pop_front() else {
-        // no more splash screens, exit
+        // There are no more splash screens, exit to the loading screen.
         next_screen.set(Screen::Loading);
         return;
     };
@@ -194,11 +194,10 @@ fn check_splash_timer(
     let container = containers.single();
     commands
         .entity(container)
-        .despawn_descendants() // previous splash images
+        .despawn_descendants()
         .with_children(|parent| {
             parent.spawn(splash_image_bundle(next_splash_image));
         });
 
-    // reset the timer
     timer.0.reset();
 }
--- a/src/screens/splash.rs
+++ b/src/screens/splash.rs
@@ -2,11 +2,7 @@
 
 use std::collections::VecDeque;
 
-use bevy::{
-    input::common_conditions::input_just_pressed,
-    prelude::*,
-    render::texture::{ImageLoaderSettings, ImageSampler},
-};
+use bevy::{input::common_conditions::input_just_pressed, prelude::*};
 
 use super::Screen;
 use crate::{theme::prelude::*, AppSet};
@@ -14,6 +10,9 @@ use crate::{theme::prelude::*, AppSet};
 pub(super) fn plugin(app: &mut App) {
     // Spawn splash screen.
     app.insert_resource(ClearColor(SPLASH_BACKGROUND_COLOR));
+
+    app.register_type::<SplashScreenContainer>();
+    app.register_type::<SplashScreenImageList>();
     app.init_resource::<SplashScreenImageList>();
 
     app.add_systems(OnEnter(Screen::Splash), spawn_splash);
@@ -53,17 +52,22 @@ const SPLASH_BACKGROUND_COLOR: Color = Color::srgb(0.157, 0.157, 0.157);
 const SPLASH_DURATION_SECS: f32 = 1.8;
 const SPLASH_FADE_DURATION_SECS: f32 = 0.6;
 
-#[derive(Component)]
+#[derive(Component, Reflect)]
+#[reflect(Component)]
 struct SplashScreenContainer;
 
-#[derive(Resource)]
-struct SplashScreenImageList(VecDeque<&'static str>);
+#[derive(Resource, Reflect)]
+#[reflect(Resource)]
+struct SplashScreenImageList(VecDeque<Handle<Image>>);
 
-impl Default for SplashScreenImageList {
-    fn default() -> Self {
-        // Use paths here rather than an ImageHandle as the loading screen hasn't
-        // run yet. To show your own splash images, replace or add here
-        Self(VecDeque::from_iter(["images/splash.png"]))
+impl FromWorld for SplashScreenImageList {
+    fn from_world(world: &mut World) -> Self {
+        let asset_server = world.resource::<AssetServer>();
+
+        // To show your own splash images, replace or add images here
+        Self(VecDeque::from_iter(
+            [asset_server.load("images/splash.png")],
+        ))
     }
 }
 
@@ -71,7 +75,9 @@ fn exit_splash_screen(mut next_screen: ResMut<NextState<Screen>>) {
     next_screen.set(Screen::Loading);
 }
 
-fn splash_image_bundle(asset_server: &AssetServer, path: &'static str) -> impl Bundle {
+fn splash_image_bundle(image: Handle<Image>) -> impl Bundle {
+    let image = image.into();
+
     (
         Name::new("Splash image"),
         ImageBundle {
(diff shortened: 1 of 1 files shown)


src/screens/splash.rs
```diff
@@ -35,48 +40,84 @@ pub(super) fn plugin(app: &mut App) {
...
-fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
-    commands
-        .ui_root()
-        .insert((
-            Name::new("Splash screen"),
+#[derive(Component)]
+struct SplashScreenContainer;
+
+#[derive(Resource)]
+struct SplashScreenImageList(VecDeque<&'static str>);
```
benfrankel, line 60: This can be a `Vec` instead of a `VecDeque`.

src/screens/splash.rs
```diff
@@ -35,48 +40,84 @@ pub(super) fn plugin(app: &mut App) {
...
+
+impl Default for SplashScreenImageList {
+    fn default() -> Self {
+        // Use paths here rather than an ImageHandle as the loading screen hasn't
+        // run yet. To show your own splash images, replace or add here
+        Self(VecDeque::from_iter(["images/splash.png"]))
+    }
+}
```
benfrankel, line 68: Can `impl FromWorld` instead, to get access to `world.resource::<AssetServer>()`. Then loading the images can be done here.

src/screens/splash.rs
```diff
@@ -35,48 +40,84 @@ pub(super) fn plugin(app: &mut App) {
...
 const SPLASH_BACKGROUND_COLOR: Color = Color::srgb(0.157, 0.157, 0.157);
 const SPLASH_DURATION_SECS: f32 = 1.8;
 const SPLASH_FADE_DURATION_SECS: f32 = 0.6;
 
-fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
-    commands
-        .ui_root()
-        .insert((
-            Name::new("Splash screen"),
+#[derive(Component)]
```
benfrankel, line 56:
```suggestion
#[derive(Component, Reflect)]
#[reflect(Resource)]
```

src/screens/splash.rs
```diff
@@ -35,48 +40,84 @@ pub(super) fn plugin(app: &mut App) {
...
+struct SplashScreenContainer;
+
+#[derive(Resource)]
```
benfrankel, line 59:
```suggestion
#[derive(Resource, Reflect)]
#[reflect(Resource)]
```

src/screens/splash.rs
```diff
@@ -135,8 +176,34 @@ fn tick_splash_timer(time: Res<Time>, mut timer: ResMut<SplashTimer>) {
...
+    mut next_screen: ResMut<NextState<Screen>>,
+    mut splash_screens: ResMut<SplashScreenImageList>,
+    containers: Query<Entity, With<SplashScreenContainer>>,
+) {
+    if !timer.0.just_finished() {
+        return;
+    }
+
+    // try to get  the next splash screen. If there isn't one, we move on to the next screen
+    if let Some(next_splash_image) = splash_screens.0.pop_front() {
```
benfrankel, line 192: `let-else` then set the next screen and `return` here to remove one level of indentation.

src/screens/splash.rs
```diff
@@ -135,8 +176,34 @@ fn tick_splash_timer(time: Res<Time>, mut timer: ResMut<SplashTimer>) {
...
+        // despawn other splash images
+        if let Ok(container) = containers.get_single() {
```
benfrankel, line 194: `let-else` here as well.

src/screens/splash.rs
benfrankel, line 1: Comments should begin with a capital letter and end with a `.` for style consistency.

benfrankel (changes requested):

src/screens/splash.rs
benfrankel: Also make sure to add `app.register_type` to `plugin`.

src/screens/splash.rs
benfrankel: Also make sure to add `app.register_type` to `plugin`.

src/screens/splash.rs
will-hart: I may be missing something but I feel like `VecDeque` is the better option here as we're popping from the front. 

There is `Vec::pop` which pops from the back, meaning we'd show the images in reverse order. The docs for `pop` actually say:

> If you'd like to pop the first element, consider using VecDeque::pop_front instead.

or there is something like `Vec::drain` but that returns an iterator so there is a lot more boilerplate.

benfrankel: That's true. We would either need to store the images in reverse order, use `vec.remove(0)` (when we know the vec is non-empty), or use an index instead of removing the images from the list. The index could reset to 0 when entering `Screen::Splash`.

I normally don't see `VecDeque` unless it's a performance optimization, which isn't relevant in this case.

src/screens/splash.rs
will-hart: Thanks - I forgot there is `let-else` 😁 

The container *should* exist, so what do you think about just doing `get()` here?

src/screens/splash.rs
benfrankel: I'll defer to other opinions on this.

src/screens/splash.rs
benfrankel: I'm an `unwrap` hater so I'd say just do `let-else`, and in my own code I would do `let container = r!(containers.get_single());` using the shorthand `or_return!` macro from `tiny_bail`.

benfrankel: Oh actually after a second look, I would rename `containers` -> `container_query` and make this a `for` loop. No need to assume / check / whatever that there's exactly one matching entity and special-case this, better to handle multiple entities for free.

src/screens/splash.rs
will-hart: Yep, happy to change but this is exactly the kind of thing `VecDeque` is in `std::collections` for.

src/screens/splash.rs
```diff
@@ -135,8 +173,31 @@ fn tick_splash_timer(time: Res<Time>, mut timer: ResMut<SplashTimer>) {
...
+    mut splash_images: ResMut<SplashScreenImageList>,
+    containers: Query<Entity, With<SplashScreenContainer>>,
+) {
+    if !timer.0.just_finished() {
+        return;
     }
+
+    // Try to get the next splash image. If there isn't one, we move on to the next screen.
+    let Some(next_splash_image) = splash_images.0.pop_front() else {
+        // There are no more splash screens, exit to the loading screen.
```
benfrankel, line 189:
```suggestion
    // Try to get the next splash image.
    let Some(next_splash_image) = splash_images.0.pop_front() else {
        // There are no more splash images, so continue to the loading screen.
```

benfrankel (changes requested):

src/screens/splash.rs
```diff
@@ -35,48 +39,82 @@ pub(super) fn plugin(app: &mut App) {
...
+#[reflect(Resource)]
+struct SplashScreenImageList(VecDeque<Handle<Image>>);
+
+impl FromWorld for SplashScreenImageList {
+    fn from_world(world: &mut World) -> Self {
+        let asset_server = world.resource::<AssetServer>();
+
+        // To show your own splash images, replace or add images here.
+        Self(VecDeque::from_iter(
+            [asset_server.load("images/splash.png")],
```
benfrankel, line 69: This should keep the original `load_with_settings` and comments from before the PR.

benfrankel (changes requested):

src/screens/splash.rs
```diff
@@ -135,8 +173,34 @@ fn tick_splash_timer(time: Res<Time>, mut timer: ResMut<SplashTimer>) {
...
+    };
+
+    commands
+        .entity(container)
+        .despawn_descendants()
+        .with_children(|parent| {
+            parent.spawn(splash_image_bundle(next_splash_image));
+        });
+
+    timer.0.reset();
```
benfrankel, line 205: Change `TimerMode::Once` -> `TimerMode::Repeating`, and then this line isn't needed.

benfrankel (changes requested):

src/screens/splash.rs
```diff
@@ -35,48 +39,82 @@ pub(super) fn plugin(app: &mut App) {
...
+        },
+    )
+}
+
+fn spawn_splash(
+    mut commands: Commands,
+    splash_images: Res<SplashScreenImageList>,
+    mut next_screen: ResMut<NextState<Screen>>,
+) {
+    if splash_images.0.is_empty() {
```
janhohenheim, line 106: Don't think this check is necessary. Doesn't the system with the timer do this for us anyways?

src/screens/splash.rs
```diff
@@ -35,48 +39,82 @@ pub(super) fn plugin(app: &mut App) {
...
 const SPLASH_FADE_DURATION_SECS: f32 = 0.6;
 
-fn spawn_splash(mut commands: Commands, asset_server: Res<AssetServer>) {
-    commands
-        .ui_root()
-        .insert((
-            Name::new("Splash screen"),
+#[derive(Component, Reflect)]
+#[reflect(Component)]
+struct SplashScreenContainer;
```
janhohenheim, line 57: Add a doc comment saying why we need this and what it represents.

janhohenheim (changes requested):
Since this veers into the territory of "nice to have, but not necessary for most users", I want @alice-i-cecile's opinion on whether we should add this. I'm personally ambivalent.

In any case, consider upstreaming the code as an official example. I think it would be really helpful there :D

src/screens/splash.rs
alice-i-cecile: Keep it as vec-deque but add a comment about why it's used IMO.

alice-i-cecile:
I like both of these, but strongly seconding the recommendation to upstream the splash screen code as an official example in the `games` folder!

src/screens/splash.rs
will-hart: This has been removed, but this prevents a pause for the timer's duration with a blank screen if no images are provided.

will-hart:
OK, I've removed the multi-splash screen code, resolved any discussions relating that and updated the PR description.

benfrankel (approved):

src/screens/splash.rs
```diff
@@ -35,12 +36,23 @@ pub(super) fn plugin(app: &mut App) {
         )
             .run_if(in_state(Screen::Splash)),
     );
+
+    // Exit the splash screen early if the player hits escape.
+    app.add_systems(
+        Update,
+        exit_splash_screen
+            .run_if(input_just_pressed(KeyCode::Escape).and_then(in_state(Screen::Splash))),
```
janhohenheim, line 44: I personally think that this reads better when written as two separate `run_if` because of the line break, but I'm not blocking on that.

janhohenheim (approved):

src/screens/splash.rs
benfrankel: IMO that's confusing because you have to know whether that yields an `and` or an `or` behavior.

Issue #220, opened 2024-08-04
Text on credits screen is awkwardly justified


rparrett:
Text that flows onto multiple lines is left justified, but text that is one line only gets centered.

We should either add `with_text_justify(JustifyText::Center)` to the text, or remove `justify_content: JustifyContent::Center` from the label container.

<img width="1280" alt="image" src="https://github.com/user-attachments/assets/dd8d4da1-3468-44e5-a8bf-d1729e42042a">

benfrankel:
Alternatively, a grid layout could be nice (2 columns: right-aligned names, left-aligned descriptions).

benfrankel:
This is resolved by https://github.com/TheBevyFlock/bevy_quickstart/pull/259. Better UI layouting will come in later PRs.

PR #259, opened 2024-08-10
Fix label alignment

janhohenheim:
Trivial fix for #220 
We can improve this further in future PRs, but this is at least better for the moment.

image

### DIFF
--- a/src/screens/loading.rs
+++ b/src/screens/loading.rs
@@ -22,7 +22,10 @@ fn show_loading_screen(mut commands: Commands) {
         .ui_root()
         .insert(StateScoped(Screen::Loading))
         .with_children(|children| {
-            children.label("Loading...");
+            children.label("Loading...").insert(Style {
+                justify_content: JustifyContent::Center,
+                ..default()
+            });
         });
 }
 
--- a/src/theme/widgets.rs
+++ b/src/theme/widgets.rs
@@ -98,7 +98,7 @@ impl<T: Spawn> Widgets for T {
             NodeBundle {
                 style: Style {
                     width: Px(500.0),
-                    justify_content: JustifyContent::Center,
+                    justify_content: JustifyContent::FlexStart,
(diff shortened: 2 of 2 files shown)


src/theme/widgets.rs
```diff
@@ -98,7 +98,7 @@ impl<T: Spawn> Widgets for T {
             NodeBundle {
                 style: Style {
                     width: Px(500.0),
-                    justify_content: JustifyContent::Center,
+                    justify_content: JustifyContent::FlexStart,
```
benfrankel, line 101: Fully deleting this line should work too I believe?

benfrankel (changes requested):

src/theme/widgets.rs
janhohenheim: Oh, right. On it

janhohenheim: [`2a6d9dc` (#259)](https://github.com/TheBevyFlock/bevy_quickstart/pull/259/commits/2a6d9dceef88144b4879a06ca64dcfc34f8c2208)

Also fixed the loading screen not being centered.

src/screens/loading.rs
```diff
@@ -22,7 +22,10 @@ fn show_loading_screen(mut commands: Commands) {
         .ui_root()
         .insert(StateScoped(Screen::Loading))
         .with_children(|children| {
-            children.label("Loading...");
+            children.label("Loading...").insert(Style {
+                justify_content: JustifyContent::Center,
+                ..default()
+            });
```
benfrankel, line 28: I don't think this does anything? This overwrites the initial value of:
```rust

                style: Style {
                    width: Px(500.0),
                    justify_content: JustifyContent::Center,
                    align_items: AlignItems::Center,
                    ..default()
                },
```
Shouldn't that already be centering the text?

benfrankel (approved):

PR #260, opened 2024-08-10
Refactor label

janhohenheim:


benfrankel:
Repeat this for the header widget?

janhohenheim:
Exactly, the bg color would no longer work the same

benfrankel (approved):
Approving under the assumption that you checked that everything looks fine.

PR #263, opened 2024-08-12
Automate `cargo-generate` branch updates

### LINKED ISSUE #129: Automatically merge `main` into `cargo-generate` when a PR is merged
Merge conflict could automatically open an issue.
If this is done via workflow, make sure it is excluded from `cargo-generate.toml` on the `cargo-generate` branch.
Assigning the local CI wizard ;p

BD103:
Closes #129.

This adds `cargo-generate.yaml`, a workflow that will automatically merge new changes from `main` to the `cargo-generate` branch.

This will require someone with write access to this repo to create a personal access token with write permissions for "Contents" and "Workflows". I cannot use my own, since I don't have write access. I recommend creating a scoped token that can only write to this specific repository.

This workflow will automatically create an issue on a merge conflict. For an example, see https://github.com/BD103/bevy_quickstart/issues/4.


.github/workflows/cargo-generate.yaml
```diff
@@ -0,0 +1,110 @@
+# This workflow is used to automatically update the `cargo-generate` branch. If you copied the
+# template, you can safely delete this. :)
+#
+# BE VERY CAREFUL about what this workflow runs, since IT HAS THE ABILITY TO DELETE THE ENTIRE
+# REPOSITORY'S CONTENTS. For reason, stick to pre-installed tools and trusted actions (such as
```
janhohenheim, line 5:
```suggestion
# REPOSITORY'S CONTENTS. For this reason, stick to pre-installed tools and trusted actions (such as
```

.github/workflows/cargo-generate.yaml
```diff
@@ -0,0 +1,110 @@
...
+# CARGO_GENERATE_PAT.
+permissions:
+  # We need write permissions to create new issues.
+  issues: write
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    # Only run this job on the source repository, skipping forks.
+    if: ${{ github.repository == 'TheBevyFlock/bevy_quickstart' }}
```
janhohenheim, line 27: Really good idea!

janhohenheim (commented):
Appreciate all the comments in the workflow :)

.github/workflows/cargo-generate.yaml
BD103: Yeah! I'm not sure if it's possible to exclude this file from users of this template, so this is the safeguard.

BD103:
This is what my token looks like:

image

Save value as `CARGO_GENERATE_PAT` under this repository's actions secrets:

image

.github/workflows/cargo-generate.yaml
benfrankel: Maybe `cargo-generate.yaml` can be put in the `cargo-generate` branch? I'm not sure if it would still trigger properly on commits to `main` though.

BD103: It wouldn’t, unfortunately, though I like where your head is at!

In the end, it’s probably not too big of a deal. I put comments saying users can safely delete it, and it won’t run anything anyways on other repos.

.github/workflows/cargo-generate.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+            <details>
+              <summary><code>git diff --diff-filter=U</code></summary>
+
+              ```diff
+              GIT_DIFF_IDENTIFIER
+              ```
+            </details>
+
+            _This is an automated message created using `cargo-generate.yaml`._
```
benfrankel, line 91:
```suggestion
            _This is an automated message created by `cargo-generate.yaml`._
```
Optional.

.github/workflows/cargo-generate.yaml
```diff
@@ -0,0 +1,110 @@
...
+          token: ${{ secrets.CARGO_GENERATE_PAT }}
+
+      - name: Configure Git user
+        env:
+          # From https://github.com/actions/checkout/pull/1707.
+          NAME: github-actions[bot]
+          EMAIL: 41898282+github-actions[bot]@users.noreply.github.com
+        run: |
+          git config user.name $NAME
+          git config user.email $EMAIL
```
benfrankel, line 46:
```suggestion
          git config user.name "${NAME}"
          git config user.email "${EMAIL}"
```

benfrankel (changes requested):
The double quotes / curly braces are technically not required here, but it's good form to use them anyways to avoid subtle bugs in situations where they _are_ required, or if assumptions are invalidated in the future.

PR seems reasonable to me otherwise. We can test it out in practice and fix issues if any arise.

janhohenheim (approved):
LGTM other than @benfrankel's comments

benfrankel (approved):

PR #264, opened 2024-08-14
Remove `audio` + `assets` modules in favor of raw Bevy APIs and thin LoadResource API

cart:
I think we shouldn't encourage something like an  `assets` module containing generic things like ImageHandles, SfxHandles, BgmHandles, etc. Mixing a "footstep" sound and a "button press" sound under the same categorization is (1) not scalable to large numbers of things and (2) a missed opportunity to organize things based on context. Put "button sounds" next button logic, inside a button module. Put "footstep sounds" next to footstep logic.

The whole `audio` module (PlaySfx, PlayBgm abstractions) should be removed ... too opinionated and too centralized. Triggering clips with their string name is a pattern we should not encourage. We should just use the Bevy APIs (load handles and spawn bundles to play audio). They are more than good enough, and I want people learning how to write Bevy code, not bevy-quickstart code.

To replace the load tracking logic, I've added a "load resource" system, which allows us to define and load Resources as Assets. This allows us to rely on Bevy's built-in asset tracking to determine when the Resource is ready to be inserted. Using this pattern is preferable because it is much closer to what we will encourage after "high level asset-driven state transitions" are implemented.


janhohenheim (approved):
Thanks! Some first thoughts:
- I have no big thoughts on the asset organization. Personally, I have enjoyed knowing exactly which file I needed to look at if I wanted to change an asset, but I'm sure I would be happy with your solution as well after playing around with it 🙂
- I don't have much personal preference for the audio, other than that I would prefer having a central bottleneck through which all SFX flows as I can then easily control things like "how many sound effects are allowed to play at once?". But such an abstraction is fairly easy to build yourself, so I wouldn't block on that. I however heavily suspect that @alice-i-cecile will disagree with the lack of audio abstraction here as this version comes directly from her design. In fact, since this API is coming from here PR to the Bevy examples, this is closer to Bevy code than to bevy_quickstart code 😅
- If the asset loading can be made to resemble the future official solution, I'm all for it! 😄

Approving the PR since I have nothing against it really. Will wait with merging until others have bikeshed it.

src/screens/playing.rs
```diff
@@ -16,13 +17,39 @@ pub(super) fn plugin(app: &mut App) {
     );
 }
 
-fn spawn_level(mut commands: Commands) {
+#[derive(Resource, Asset, Reflect, Clone)]
+pub struct GameplayMusic {
```
alice-i-cecile, line 21: I don't like the duplication of structures here. This really feels like it should be a single resource synced to the game state. Only one track should be playing at once.

alice-i-cecile:
Overall opinions:

1. I really like the `LoadResource` trait. I still wish we had first-party solutions for "wait for everything to load", but this is a really nice use of more sophisticated asset features.
2. I definitely dislike splitting music into multiple resources: you need to manage things like "what track is playing" and handle cross-fading in a single coherent spot (and the code duplication is eh).
3. Moving away from string labels is fine: ultimately you want this to either be enums or data-driven.
4. I have similar complaints about scattering sound effects throughout the codebase. Cohesively doing things like adding spatialization, setting the volume, blending and so on becomes much harder. I think that the *assets* should live with the content domain that they're part of (player sfx with player sprites for example), but more advanced features require looking at sound effects as a whole :)
5. Reducing complexities / abstractions is good!

alice-i-cecile (approved):
Overall I'm not opposed to this: it's simpler, and well-suited to a jam. I still think it's harder to polish up, but that's fine in this context.

cart:
> other than that I would prefer having a central bottleneck through which all SFX flows

In general you all are describing a Bus API: context-specific mixable audio targets (ex: A "background music" bus, a "sound effects" bus, etc). This is definitely something Bevy should support, but I will assert that pretty much any abstraction we try to build in bevy-quickstart will be:

1. Incomplete: You mention crossfading, but I'm not seeing any crossfading in the API. If you want to crossfade, you still need to drop down to the individual clips API, which leads me to the next problem.
2. Lossy: by abstracting out the Bevy Audio API in places like PlayBgm and PlaySfx (which serve kind of like pseudo-busses), you have removed Entity from the user-facing playback API. This makes it harder to do context-specific operations directly on the clip (ex: increase / decrease volume, adjust speed, etc). Now the user needs to somehow search for the Entity (aka _know_ that these operations will implicitly spawn an entity, _know_ that you've automatically inserted an IsBgm component, and then they need query for it and hope theres only one). `play_bgm()` also has implicit behaviors like despawning all currently playing clips. This fully prevents crossfade operations.

I will assert that this is not the platform to be designing such an API. The experts haven't been engaged and the oversight here is too low.

In general, I would like to (again) urge everyone to be conservative when it comes to abstraction building. We should favor doing things "the bevy way" and drive improvements upstream when things are lacking. Adding new abstractions to the template should be a last resort. They should be very scoped and augment the "bevy" way of doing things.

I _do_ agree that it makes sense to be able to categorize and control playback of audio. I think we already have a good tool for that ... marker components (which yall are already making use of in `PlayBgm`!). This is the pattern we already recommend in our examples:

```rust
fn play(mut commands: Commands, player_assets: Res<PlayerAssets>) {
    commands.spawn((
        PlayerBundle {
            source: player_assets.step().clone(),
            settings: PlaybackSettings::DESPAWN,
        },
        SoundEffect,
    ))
}

fn adjust_sound_effect_volume(sinks: Query<&AudioSink, With<SoundEffect>) {
    for sink in &sinks {
        sink.set_volume(0.5);
    }
}
```

I think this is already significantly preferable to the proposed APIs.

> I definitely dislike splitting music into multiple resources: you need to manage things like "what track is playing" and handle cross-fading in a single coherent spot (and the code duplication is eh).

This is going to happen no matter what in the context of scenes, which will each have their own audio that will be loaded and unloaded as the scene is loaded and unloaded. I also think that is "good design" generally (with or without scenes), as it allows people to build cohesive / standalone sets of functionality that nest into a larger whole. 

I will assert that constructing a thing that knows every possible piece of music that could play ahead of time prevents modularization for very little benefit. We can answer questions like "what is playing" and "how do we cross-fade" without having an uber-struct. In fact, I consider those problems to be fully independent of "where do handles live". Finding what is currently playing is a question for the live data on the spawned entities, not an uber-handle-manager-resource. Any view you try to build there will be incomplete and risks being out of date. Ditto for cross fading (which includes answering the question "what is playing" and then adjusting the volume per-frame between old and new audio entities). If/when/how these things happen is _context specific_. Any abstraction you try to build over them has a high risk of not being the right fit. Hiding these mechanisms from users isn't doing them any favors.

Again, I think marker components and raw Bevy Audio APIs are the move here.

cart:
I've added marker components, per the discussion above.

src/audio.rs
```diff
@@ -0,0 +1,7 @@
+use bevy::prelude::*;
+
+#[derive(Component, Default)]
+pub struct BackgroundMusic;
+
+#[derive(Component, Default)]
+pub struct SoundEffect;
```
janhohenheim, line 7: These are really nice! I'd appreciate some comments guiding users why they should care about them and what they might be used for. Especially a `Trigger<OnAdd, SoundEffect>` and similar would imo be quickly useful.

janhohenheim (approved):
That sounds fair enough. Thanks for the explanation. The marker components indeed would make something like "play only 25 SFX at the same time" or "generally crossfade between all BGM" easy to implement: just add an observer for `Trigger<OnAdd, SoundEffect>`.
I'll leave some minor comments, but I fully agree with the PR now :)

We actually already went from `Sfx` to `SoundEffects` and then back again to `Sfx` because we noticed that the extra wording ended up making the code more verbose but not more clear. This is the reason why we are using `Bgm` over `Soundtrack`: it's a bit less correct, but way smaller typographically. I would like to see one of the following two changes:
- Change `SoundEffect` to `Sfx` and `BackgroundMusic` to `Bgm`, or
- change `BackgroundMusic` to `Soundtrack`

src/asset_tracking.rs
```diff
@@ -0,0 +1,57 @@
...
+                    world.insert_resource(value.clone());
+                }
+            }));
+        self
+    }
+}
+
+#[derive(Resource, Default)]
+struct ResourceHandles {
+    waiting: VecDeque<(UntypedHandle, fn(&mut World, &UntypedHandle))>,
```
janhohenheim, line 38: I think `fn(&mut World, &UntypedHandle))` looks pretty scary for a beginner. Either use a simpler type or add a comment making it a bit friendlier.

src/asset_tracking.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[allow(unused)]
+    finished: Vec<UntypedHandle>,
+}
+
+fn load_resource_assets(world: &mut World) {
+    world.resource_scope(|world, mut resource_handles: Mut<ResourceHandles>| {
+        world.resource_scope(|world, assets: Mut<AssetServer>| {
+            for _ in 0..resource_handles.waiting.len() {
```
janhohenheim, line 46:
```suggestion
            for _ in &resource_handles.waiting {
```

src/demo/player.rs
```diff
@@ -63,7 +65,7 @@ fn spawn_player(
         Name::new("Player"),
         Player,
         SpriteBundle {
-            texture: image_handles[ImageHandles::PATH_DUCKY].clone_weak(),
+            texture: player_assets.ducky.clone(),
```
janhohenheim, line 68: @alice-i-cecile I remember you explicitly wanted a `clone_weak` here. I personally would prefer the `clone` shown here, but just wanted to let you know

src/demo/player.rs
```diff
@@ -110,3 +112,40 @@ fn record_player_directional_input(
         controller.intent = intent;
     }
 }
+
+#[derive(Resource, Asset, Reflect, Clone)]
+pub struct PlayerAssets {
+    #[dependency]
```
janhohenheim, line 118: I don't think `dependency` is an attribute a newcommer is likely to know, based on the fact that I just learned about it through this PR :D 
Would appreciate a comment explaining it

src/asset_tracking.rs
cart: Not the same thing, as it holds a borrow to `resource_handles.waiting`, but we modify resource_handles.waiting in the loop.

src/screens/playing.rs
```diff
@@ -16,13 +17,42 @@ pub(super) fn plugin(app: &mut App) {
...
+            ))
+            .id(),
+    );
 }
 
-fn stop_bgm(mut commands: Commands) {
-    commands.stop_bgm();
+fn stop_bgm(mut commands: Commands, mut music: ResMut<GameplayMusic>) {
+    if let Some(entity) = music.entity.take() {
+        commands.entity(entity).despawn();
```
janhohenheim, line 54:
```suggestion
        commands.entity(entity).despawn_recursive();
```
No real difference in this case, but I consider `despawn` to be a footgun as adding a child is possible from any place due to Bevy's decentralized nature

src/theme/interaction.rs
```diff
@@ -1,16 +1,22 @@
...
     app.add_systems(
         Update,
         (
             trigger_on_press,
             apply_interaction_palette,
             trigger_interaction_sfx,
-        ),
+        )
+            .run_if(not(in_state(Screen::Loading)))
+            .run_if(not(in_state(Screen::Splash))),
```
janhohenheim, line 19: Previous code has explicitly not used chained `run_if`s as that makes it unclear whether we are AND or ORing. I *personally* like chaining them, but I want the code to be consistent and I definitely see the argument here.

src/demo/player.rs
cart: In theory, this is fine because we're holding the handle elsewhere. In practice, I don't want to encourage users to clone_weak() in these cases because we really do want to tie the lifetime of the audio asset to the clips playing it. The cost of incrementing an Arc is chump-change in this context. Users that are aware of the tradeoffs here are welcome to optimize, but this is aggressively premature optimization _in the context of educating new bevy users_.

src/screens/playing.rs
cart: Reasonable!

src/theme/interaction.rs
cart: I personally think this reads better than the `and_then` chain, but I don't have strong opinions here.

src/asset_tracking.rs
janhohenheim: You're right. Changed my comment while you were typing :D

src/theme/interaction.rs
janhohenheim: @benfrankel using De Morgan, both 
```rust
.run_if(not(in_state(Screen::Loading).and_then(in_state(Screen::Splash)))
```
and 
```rust
.run_if(not(in_state(Screen::Loading)).and_then(not(in_state(Screen::Splash)))
```
seem very unreadable to me. I'm tending heavily towards keeping it as in shown in the PR. Would that be alright for you?

src/demo/player.rs
janhohenheim: Fully agree with that assessment.

cart:
> Change SoundEffect to Sfx and BackgroundMusic to Bgm or change BackgroundMusic to Soundtrack

I think we should name these whatever we would call them in the context of a theoretical "volume control screen" in a game. I haven't seen `sfx` or `bgm` in any volume control screen (to my knowledge). I do often see "sound effects" and "music" or "background music".

I think "soundtrack" implies a more linear / intentionally listenable collection of music. I consider it to be a "more specific category" that is a subset of "background music". Aka you buy "the soundtrack" for a game. Thats not necessarily the same thing as the category of audio playing that is generally music. 

I'm actually most inclined to use `SoundEffect` and `Music`

Some prior art:

image

image

image

src/theme/interaction.rs
benfrankel: That first one should be a `.or_else`, I believe.

When I read "run if A, run if B", it sounds to me like "if A is true, run the system; if B is true, run the system", which would be OR. But the actual behavior is surprisingly "if A is NOT true, DO NOT run the system; if B is NOT true, DO NOT run the system", which is AND. That only makes sense if you already know that that's what will happen. And `and_then` exists to be used, presumably.

If the `.run_if` x2 remains, it needs a comment at least.

janhohenheim:
@cart those comparisons are a very good point for using "music" over "soundtrack", you've convinced me.
My experience from having tested both writing variants leads me to still prefer the short variants for development, but I'm ultimately fine with the long versions as well.

benfrankel:
IMO code and UI are different contexts with different requirements. My personal preference would be `sfx` and `music`.

src/theme/interaction.rs
janhohenheim: @benfrankel thanks, fixed.
I agree, a comment should at least be there, even though it is "standard Bevy behavior". The syntax is counterintuitive when you read it the first time.

src/asset_tracking.rs
cart: >  could we not simplify this by making it a vec

I don't see how that is any simpler. We'd still be doing the same operations. Might as well preserve the order.

src/theme/interaction.rs
janhohenheim: To expand on my comment about unreadableness: IME chained `run_if`s will be formatted more nicely by cargo fmt than `and_then`. If it is formatted nicely in this case, I have no criticism.

cart: I've opted for `and_then`

cart:
I've opted for `SoundEffect` and `Music`.

> IMO code and UI are different contexts with different requirements. My personal preference would be sfx and music.

While I do agree, generally, I also see no reason to have two separate concept names that then later need to be coupled in these contexts. Bevy generally favors unshortened names for things in the interest of clarity (which some exceptions for uber-highly trafficked types like Res and Mut). I think it makes sense to follow that convention here. In the world of autocomplete, those extra characters don't cost much.

src/asset_tracking.rs
janhohenheim: While the code would look the same, I think the mental burden would be smaller. The simplicity comes from not having to consider the two-tailedness of operations. IME many people consider `Vec` to be the "default" container, considering its simplicity it being the only one with a convenience macro and a list or vector often being the first data structure that programmers learn to use in a new language. As such, a usage of `VecDeque` IMO sticks out as "Using this was a very important and well-thought out decision, please pay attention to it", which is not really the case. The order being preserved here is fairly arbitrary imo, and in a world where we already merged a version using `Vec` and someone opened a PR for a `VecDeque`, I'd probably reject that PR as unnecessary.
If you feel strongly about this, we can let it be.

cart:
I think this is merge-able now.

src/theme/interaction.rs
```diff
@@ -1,16 +1,21 @@
...
+    app.load_resource::<ButtonAssets>();
     app.add_systems(
         Update,
         (
             trigger_on_press,
             apply_interaction_palette,
             trigger_interaction_sfx,
-        ),
+        )
+            .run_if(not(in_state(Screen::Loading)).and_then(not(in_state(Screen::Splash)))),
```
benfrankel, line 18: I'm curious why this run condition is needed. Ideally, `src/theme/` shouldn't depend on `src/screens/`.

src/theme/interaction.rs
```diff
@@ -57,15 +62,55 @@ fn apply_interaction_palette(
     }
 }
 
+#[derive(Resource, Asset, Reflect, Clone)]
+pub struct ButtonAssets {
```
benfrankel, line 66: I think the naming should be consistent between this and `trigger_interaction_sfx`. If not, a doc comment should be added to this resource that makes the connection clear, like `/// Interaction sound effects for buttons.`.

As things currently stand, `trigger_interaction_sfx` only requires the `Interaction` component, so it can apply to things that may not be considered buttons.

cart:
Ah forgot the dependency comment.

src/theme/interaction.rs
```diff
@@ -57,15 +62,55 @@ fn apply_interaction_palette(
...
+            Interaction::Pressed => {
+                commands.spawn((
+                    AudioBundle {
+                        source: button_assets.press.clone(),
+                        settings: PlaybackSettings::DESPAWN,
+                    },
+                    SoundEffect,
+                ));
+            }
+            _ => {}
```
benfrankel, line 113: This could be simplified to only clone the handle within the `match` block, then have a single `commands.spawn` afterwards with the chosen handle.

cart:
Also worth calling out that ultimately we'd like to have something like this:

```rust
#[derive(Resource, AssetCollection, Reflect, Clone)]
pub struct PlayerAssets {
    pub ducky: Handle<Image>,
    pub steps: Vec<Handle<AudioSource>>,
}
```

Where the `dependency` attribute is implicit. When that rolls out this code should change to use it.

src/asset_tracking.rs
cart: Reasonable. I don't feel strongly. Pushed the change!

src/asset_tracking.rs
benfrankel, line 1: A module-level doc comment + an update to the project structure table in the "Write your game" section of `README.md` would help make this module more approachable.

src/asset_tracking.rs
```diff
@@ -0,0 +1,57 @@
...
+            let assets = world.resource::<Assets<T>>();
+            if let Some(value) = assets.get(handle.id().typed::<T>()) {
+                world.insert_resource(value.clone());
+            }
+        }));
+        self
+    }
+}
+
+/// A function that inserts a loaded resource
```
benfrankel, line 33:
```suggestion
/// A function that inserts a loaded resource.
```

src/audio.rs
```diff
@@ -0,0 +1,37 @@
+use bevy::prelude::*;
+
+/// An organizational marker component that should be added to a spawned [`AudioBundle`] if it is in the
+/// general "music" category (ex: global background music, soundtrack, etc).
+///
+/// This can then be used to query for and operate on sounds in that category. For example:
+///
+/// ```
```
benfrankel, line 8: Are code blocks in doc quotes considered `rust` by default?

benfrankel (changes requested):

src/audio.rs
janhohenheim: Yep

benfrankel: Ok cool thanks :)

src/demo/animation.rs
```diff
@@ -4,12 +4,12 @@
...
 
-use std::time::Duration;
-
 use bevy::prelude::*;
+use rand::seq::SliceRandom;
+use std::time::Duration;
 
 use super::movement::MovementController;
-use crate::{assets::SfxHandles, audio::sfx::SfxCommands as _, AppSet};
+use crate::{audio::SoundEffect, demo::player::PlayerAssets, screens::Screen, AppSet};
```
benfrankel, line 12: I'm personally not a fan of `use super::`, but we should be consistent in its usage anyways. Move `crate::demo::player::PlayerAssets` to `super::player::PlayerAssets`.

(GitHub UI won't let me write a `suggestion` for this)

src/demo/animation.rs
```diff
@@ -24,6 +24,7 @@ pub(super) fn plugin(app: &mut App) {
                 trigger_step_sfx,
             )
                 .chain()
+                .run_if(in_state(Screen::Playing))
```
benfrankel, line 27: Is this run condition necessary? Maybe you want a ducky walking around your title screen.

src/audio.rs
```diff
@@ -0,0 +1,37 @@
...
+/// use bevy::prelude::*;
+/// use bevy_quickstart::audio::Music;
+///
+/// fn set_music_volume(sinks: Query<&AudioSink, With<Music>>) {
+///     for sink in &sinks {
```
benfrankel, line 13:
```suggestion
/// fn set_music_volume(sink_query: Query<&AudioSink, With<Music>>) {
///     for sink in &sink_query {
```

src/audio.rs
```diff
@@ -0,0 +1,37 @@
...
+/// general "sound effect" category (ex: footsteps, the sound of a magic spell, a door opening).
+///
+/// This can then be used to query for and operate on sounds in that category. For example:
+///
+/// ```
+/// use bevy::prelude::*;
+/// use bevy_quickstart::audio::SoundEffect;
+///
+/// fn set_sound_effect_volume(sinks: Query<&AudioSink, With<SoundEffect>>) {
+///     for sink in &sinks {
```
benfrankel, line 31:
```suggestion
/// fn set_sound_effect_volume(sink_query: Query<&AudioSink, With<SoundEffect>>) {
///     for sink in &sink_query {
```

src/demo/animation.rs
```diff
@@ -4,12 +4,12 @@
 //! - [Sprite animation](https://github.com/bevyengine/bevy/blob/latest/examples/2d/sprite_animation.rs)
 //! - [Timers](https://github.com/bevyengine/bevy/blob/latest/examples/time/timers.rs)
 
-use std::time::Duration;
-
 use bevy::prelude::*;
+use rand::seq::SliceRandom;
```
benfrankel, line 8: Optional:
```suggestion
use rand::prelude::*;
```

benfrankel (changes requested):

src/demo/animation.rs
janhohenheim: I don't like super for this either. Pretty sure existing usages were placed there by someone’s IDE and we should replace them in a follow-up PR instead

src/audio.rs
janhohenheim: Context: this is for consistency. Don’t care much about which version we ise, but we habe been consistent with _queue for now.

src/theme/interaction.rs
janhohenheim: Agreed. I'm fine with using interaction or button everywhere, just need to be consistent

src/demo/animation.rs
benfrankel: Another option is something like `.run_if(resource_exists::<PlayerAssets>)`, instead of coupling to the `Screen` state.

src/demo/animation.rs
benfrankel: If we want to replace `super` with `crate` in a separate PR, I'm fine with leaving this as is in this PR.

src/theme/interaction.rs
cart: This exists to ensure it doesn't run before the `ButtonAssets` resource (and associated assets) have been loaded. The asset loading lifecycle is currently tied to `Screen`. 

An alternative would be to turn the `Res<ButtonAssets>` in `trigger_interaction_sfx` into `Option<Res<ButtonAssets>>` to allow the system to run panic-free when the resource doesn't exist yet. However conceptually, I think tying this into the current (or a separate) asset loading lifecycle is the right move.

src/theme/interaction.rs
cart: Makes sense! I'll align the name.

src/theme/interaction.rs
cart: Good call!

src/demo/animation.rs
cart: > .run_if(resource_exists::<PlayerAssets>)

Thats a reasonable way to do it I think. That being said, I do think tying into states for determining if a resource can/should be used is a good idea generally.

src/audio.rs
cart: I'm generally in favor of "plural short names for queries" instead of tacking on `_query` to everything. But happy to align for now.

src/theme/interaction.rs
cart: I've switched to `resource_exists` here

src/screens/playing.rs
cart: Resolving this as I think we've agreed that distributed is the move here.

cart:
Comments have been resolved!

benfrankel (approved):

PR #268, opened 2024-08-15
Simplify commands a bit

### LINKED ISSUE #265: Applying commands within other commands can be done more simply
Spotted here: https://github.com/TheBevyFlock/bevy_quickstart/blob/7afa21d3f8cd471456872c5721ba7ca6e6357a70/src/demo/level.rs#L20

Instead, just use the `apply` method defined in the `Command` trait directly.

### LINKED ISSUE #266: Structs that impl Command can be replaced with bare functions
When you don't need to configure anything, you can take advantage of the blanket implementation. See https://docs.rs/bevy/latest/bevy/ecs/world/trait.Command.html#impl-Command-for-F

janhohenheim:
Resolves #265 
Resolves #266

### DIFF
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -10,13 +10,10 @@ pub(super) fn plugin(_app: &mut App) {
     // later if needed.
 }
 
-#[derive(Debug)]
-pub struct SpawnLevel;
-
-impl Command for SpawnLevel {
-    fn apply(self, world: &mut World) {
-        // The only thing we have in our level is a player,
-        // but add things like walls etc. here.
-        SpawnPlayer { max_speed: 400.0 }.apply(world);
-    }
+/// A [`Command`] to spawn the level.
+/// Exclusive systems, i.e. systems that only accept `&mut World`, implement [`Command`].
+pub fn spawn_level(world: &mut World) {
+    // The only thing we have in our level is a player,
+    // but add things like walls etc. here.
+    SpawnPlayer { max_speed: 400.0 }.apply(world);
 }
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -11,7 +11,8 @@ pub(super) fn plugin(_app: &mut App) {
 }
 
 /// A [`Command`] to spawn the level.
-/// Exclusive systems, i.e. systems that only accept `&mut World`, implement [`Command`].
+/// Functions that accept only `&mut World` as their parameter implement [`Command`].
+/// We use this style when a command requires no configuration.
 pub fn spawn_level(world: &mut World) {
     // The only thing we have in our level is a player,
     // but add things like walls etc. here.
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -17,10 +17,6 @@ impl Command for SpawnLevel {
     fn apply(self, world: &mut World) {
         // The only thing we have in our level is a player,
         // but add things like walls etc. here.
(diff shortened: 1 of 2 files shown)


src/screens/playing.rs
```diff
@@ -3,7 +3,9 @@
 use bevy::{input::common_conditions::input_just_pressed, prelude::*};
 
 use super::Screen;
-use crate::{asset_tracking::LoadResource, audio::Music, demo::level::SpawnLevel};
+use crate::{
+    asset_tracking::LoadResource, audio::Music, demo::level::spawn_level as spawn_level_command,
```
alice-i-cecile, line 7: I don't think the type alias is helpful here. A comment would be better.

src/demo/level.rs
```diff
@@ -10,17 +10,10 @@ pub(super) fn plugin(_app: &mut App) {
...
-        // The only thing we have in our level is a player,
-        // but add things like walls etc. here.
-        world.commands().add(SpawnPlayer { max_speed: 400.0 });
-
-        // Flush the commands we just added so that they are
-        // all executed now, as part of this command.
-        world.flush_commands();
-    }
+/// A [`Command`] to spawn the level.
+/// Exclusive systems, i.e. systems that only accept `&mut World`, implement [`Command`].
```
benfrankel, line 14: This is not exactly true: https://docs.rs/bevy_ecs/0.14.1/src/bevy_ecs/system/commands/mod.rs.html#1173-1175

Also I don't think this comment is necessary.

benfrankel (changes requested):

src/screens/playing.rs
janhohenheim: It's needed because we also have function called `spawn_level` here. Imo both of these "deserve" the name, and qualifying this one with `level::spawn_level` does not show what the difference between them is.

src/demo/level.rs
janhohenheim: Considering how much newcomers were struggling with custom commands, I think it would be even more confusing to have two completely different ways of defining them without explanation.

benfrankel: I thought the struggle with custom commands was that they're unergonomic which makes them look unwieldy, not confusing?

janhohenheim: Better? [`70a688d` (#268)](https://github.com/TheBevyFlock/bevy_quickstart/pull/268/commits/70a688d1cbb10aa1a112ca4abb35657e7210bcdf)

janhohenheim: @benfrankel IME everything concerning exclusive access to `World` is spooky

benfrankel: > Better? [`70a688d` (#268)](https://github.com/TheBevyFlock/bevy_quickstart/pull/268/commits/70a688d1cbb10aa1a112ca4abb35657e7210bcdf)

More accurate yes. I still think the doc comment calling the function a `['Command']` is sufficient.

src/screens/playing.rs
benfrankel: A point in favor of `level::level` :zipper_mouth_face:

src/demo/level.rs
janhohenheim: @alice-i-cecile, could you act as a tie-breaker here?

alice-i-cecile: I think erring on the side of more verbose here is good.

benfrankel (approved):

PR #277, opened 2024-08-16
Do not create a new issue on `cargo-generate` merge conflicts if one already exists

### LINKED ISSUE #276: Do not open new `cargo-generate` merge conflict issue when one exists
image

There's no point in [`cargo-generate.yaml`](https://github.com/TheBevyFlock/bevy_quickstart/blob/bdbdb9bf30ce1e179b1c314116f6cb0813547c30/.github/workflows/cargo-generate.yaml) opening a new issue about `cargo-generate` branch merge conflicts when one already exists! It should check that there are no open issues, _then_ create a new one.

BD103:
Closes #276.

This uses `gh` to query for open issues, counts them, then exits early if any exist.

### DIFF
--- a/.github/workflows/cargo-generate.yaml
+++ b/.github/workflows/cargo-generate.yaml
@@ -92,6 +92,20 @@ jobs:
           # This is required to use the `gh` CLI.
           GH_TOKEN: ${{ github.token }}
         run: |
+          # Get a list of the IDs of open issues created by this action, then count them using JQ.
+          OPEN_ISSUES="$(
+            gh issue list \
+              --search "is:issue is:open author:app/github-actions in:title $TITLE" \
+              --json id \
+              --jq '. | length'
+          )"
+
+          # If an issue already exists, do not create another.
+          if [[ $OPEN_ISSUES -gt 0 ]]; then
+            echo "An issue already exists on merge conflicts! Exiting without creating another."
+            exit 0
+          fi
+
           GIT_STATUS="$(git status)"
 
           # Filter the diff to only include files with merge conflicts.
--- a/.github/workflows/cargo-generate.yaml
+++ b/.github/workflows/cargo-generate.yaml
@@ -95,14 +95,14 @@ jobs:
           # Get a list of the IDs of open issues created by this action, then count them using JQ.
           OPEN_ISSUES="$(
             gh issue list \
-              --search "is:issue is:open author:app/github-actions in:title $TITLE" \
+              --search "is:issue is:open author:app/github-actions in:title ${TITLE}" \
               --json id \
               --jq '. | length'
           )"
 
           # If an issue already exists, do not create another.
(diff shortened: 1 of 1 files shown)


.github/workflows/cargo-generate.yaml
```diff
@@ -92,6 +92,20 @@ jobs:
...
+          # Get a list of the IDs of open issues created by this action, then count them using JQ.
+          OPEN_ISSUES="$(
+            gh issue list \
+              --search "is:issue is:open author:app/github-actions in:title $TITLE" \
+              --json id \
+              --jq '. | length'
+          )"
+
+          # If an issue already exists, do not create another.
+          if [[ $OPEN_ISSUES -gt 0 ]]; then
```
benfrankel, line 104:
```suggestion
          if [[ "${OPEN_ISSUES}" -gt 0 ]]; then
```

.github/workflows/cargo-generate.yaml
```diff
@@ -92,6 +92,20 @@ jobs:
           # This is required to use the `gh` CLI.
           GH_TOKEN: ${{ github.token }}
         run: |
+          # Get a list of the IDs of open issues created by this action, then count them using JQ.
+          OPEN_ISSUES="$(
+            gh issue list \
+              --search "is:issue is:open author:app/github-actions in:title $TITLE" \
```
benfrankel, line 98:
```suggestion
              --search "is:issue is:open author:app/github-actions in:title ${TITLE}" \
```

.github/workflows/cargo-generate.yaml
```diff
@@ -92,6 +92,20 @@ jobs:
...
+              --json id \
+              --jq '. | length'
+          )"
+
+          # If an issue already exists, do not create another.
+          if [[ $OPEN_ISSUES -gt 0 ]]; then
+            echo "An issue already exists on merge conflicts! Exiting without creating another."
```
benfrankel, line 105:
```suggestion
            echo 'An issue already exists on merge conflicts! Exiting without creating another.'
```

benfrankel (changes requested):
Small nits but looks good.

janhohenheim (approved):

BD103:
@benfrankel out of curiosity, why do you prefer to make all `$VARIABLES` as `"${VARIABLES}"`? As far as I can tell, they do the same thing.

benfrankel:
@BD103 They help disambiguate in the general case. For example:

```sh
echo "$FOO-bar"
echo "${FOO}-bar"
echo "${FOO-bar}"
```

Ofc if you're just doing `"$FOO"`, there's no disambiguation needed, but in that case I still prefer `"${FOO}"` for the consistency and so that I never have to even think "do I need `{}` here?", I just always add them.

benfrankel (approved):

PR #282, opened 2024-08-17
Fix tiny typo, `screen` -> `screens`

victorb:
Currently links to

https://github.com/TheBevyFlock/bevy_quickstart/blob/main/src/screen/credits.rs

but should be

https://github.com/TheBevyFlock/bevy_quickstart/blob/main/src/screens/credits.rs


benfrankel (approved):

benfrankel:
Thanks!

Issue #285, opened 2024-08-26
Windows build failures: can't find built executable


richchurcher:
I can't tell if I've unearthed a problem or just displaying a PEBCAK! Posted to Discord, but maybe this is a better place in hindsight...

My CI seems to be failing due to lack of an executable name. Hopefully typing this out will rubber-duck a solution, but this is what I'm seeing:
```
    Finished `release-native` profile [optimized] target(s) in 28m 14s
mv: cannot stat 'target/x86_64-pc-windows-msvc/release-native/.exe': No such file or directory
Error: Process completed with exit code 1.
```
This particular game was a manual repo rather than using `cargo generate`, and I imagine I've accidentally missed a step. However, there are no instances of `bevy_quickstart` remaining in the repo (at least according to `rg`) and the `release.yaml` appears to have all the correct replacements made.

It seems the Windows build uses:
```
        run: Compress-Archive -Path '${{ env.app_package_name }}' -DestinationPath '${{ env.package }}'
```
and these values appear to be available:
```
Run for target in x86_64-pc-windows-msvc; do
  for target in x86_64-pc-windows-msvc; do
    cargo build --profile='release-native' --target="${target}" --no-default-features --features=''
    mv target/"${target}"/'release-native/.exe' tmp/binary/"${target}"'.exe'
  done
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
  env:
    cargo_build_binary_name: either
    assets_path: assets
    upload_to_github: true
    upload_to_itch: eloquentgeek/either
    app_id: games.eloquentgeek.either
    app_binary_name: either
    app_package_name: either
    app_display_name: Either
    app_short_name: Either
    git_lfs: false
    version: v0.14.0
    app: tmp/app/either
    package: either-windows.zip
    CARGO_HOME: C:\Users\runneradmin/.cargo
    CARGO_INCREMENTAL: 0
    CARGO_TERM_COLOR: always
    workflow_path: .github/workflows/release.yaml
```
The web and Linux builds work fine. Unless `either` has some kind of special meaning in the PowerShell world? :slightly_smiling_face:

richchurcher:
Will close and document here if someone on Discord lets me know what I'm not seeing!

richchurcher:
Thanks Ben! (Turns out just a tweak to a variable name after a refactor was required.)

janhohenheim:
@vineet1401 I deleted your comment because it looks like spam.

PR #286, opened 2024-08-26
Fix release workflow

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/285.


richchurcher (approved):
:shipit:

janhohenheim (approved):
Good catch

Issue #212, opened 2024-08-01
The CI uses up a lot of resources


TheKnarf:
![Screenshot 2024-08-01 at 14 41 38](https://github.com/user-attachments/assets/e867ab2a-9240-4409-bdb2-019bf9220da4)

A single game jam was all it took.

Too late I realized that I should disable the windows/linux/mac build since I only really needed the web build for the game jam. I think the template probably should disable those by default and require people to enable them if they need them.

Maybe there are other things one could do to reduce the amount of minutes it takes up? Maybe running `cargo fmt` and `cargo test` isn't necessary by default on main. Maybe one could run `cargo test` without triggering it parsing the docs folder? Are there other improvements one could make?

TheKnarf:
![Screenshot 2024-08-01 at 14 47 00](https://github.com/user-attachments/assets/e2470ed6-0b80-4ea0-a6db-b96c0b2a91fc)

CI builds took almost 30 minutes to run.

![Screenshot 2024-08-01 at 14 47 39](https://github.com/user-attachments/assets/93f2afbf-2d53-4f49-a007-0dafbaca2e08)

Release build took over 30 minutes when building for all targets, less than 15 minutes after disabling everything except for the web build.

janhohenheim:
We intentionally made it do as much stuff as possible, assuming that most users would open source their code by default (which means infinite GitHub actions) or be ready to deal either the consequences of having a private repository, i.e. fiddling with the workflow.

That said, it would be fairly easy to enable or disable the builds based on a `cargo generate` flag. How would you feel about that?

The builds should take like a minute at most when cached. Did something go wrong there?

TimJentzsch:
Under the hood it uses the GitHub Actions cache, I'm not aware of any restrictions for private repositories.

benfrankel:
Yeah I'm curious if caching wasn't working, or otherwise why it was taking so long. Looking at my own game's CI, I see this error when trying to save the cache (https://github.com/benfrankel/blobo_party/actions/runs/10139032444):

```
zstd: error 70 : Write error : cannot write block : No space left on device 
/usr/bin/tar: cache.tzst: Cannot write: Broken pipe
/usr/bin/tar: Child returned status 70
/usr/bin/tar: Error is not recoverable: exiting now
Warning: Failed to save: "/usr/bin/tar" failed with error: The process '/usr/bin/tar' failed with exit code 2
```

[...]

PR #289, opened 2024-08-28
Enable cache sweeping in CI

BD103:
Addresses part of #212.

I recently merged a feature into [Leafwing-Studios/cargo-cache](https://github.com/Leafwing-Studios/cargo-cache) v2.3.0 that automatically deletes unused build artifacts before creating a new cache, to prevent cache sizes from snowballing.

It's disabled by default due to backward compatibility reasons, so this PR enables it. Cache sweeping should keep cache sizes small and workflow runs fast.

(As an side, should a cache be used at all in the release build? Caching prevents builds from being reproducible, and can introduce the occasional issue when building.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0 OR CC0-1.0"
 
 [dependencies]
-bevy = { version = "0.14", features = ["wayland"] }
+bevy = { version = "0.14", features = ["x11"] }
 rand = "0.8"
 # Compile low-severity logs out of native builds for performance.
 log = { version = "0.4", features = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0 OR CC0-1.0"
 
 [dependencies]
-bevy = { version = "0.14", features = ["x11"] }
+bevy = { version = "0.14", features = ["wayland"] }
 rand = "0.8"
 # Compile low-severity logs out of native builds for performance.
 log = { version = "0.4", features = [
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0 OR CC0-1.0"
 
 [dependencies]
-bevy = { version = "0.14", features = ["wayland"] }
```
benfrankel, line 8: Why remove `wayland` support? Also, `x11` is a default feature, so no need to specify it explicitly.

benfrankel (commented):
Caching in release is nice for faster iteration during a game jam (e.g. with playtesters). There could be a flag in the env variables to disable it, but it should be supported and probably enabled by default.

Cargo.toml
BD103: It's a temporary change to prove that this PR works! I'm reverting 7188e2c623e4b4fc97cd9abad3d55c3ae364544d after the workflow finishes :sweat_smile:

benfrankel: Right that makes total sense :)

janhohenheim (approved):
Agreed with @benfrankel. I think the pros outweigh the cons for most people, and if you care about these things enough, you can disable the cache very simply.

BD103:
I made a temporary commit (7188e2c623e4b4fc97cd9abad3d55c3ae364544d) that updates some dependencies, which will make some build artifacts stale. After looking at [the run](https://github.com/TheBevyFlock/bevy_quickstart/actions/runs/10605969858), it removed 365 MiB on the Clippy and Docs jobs and removed a whopping 3.68 GiB on the Tests job.

These savings will only occur when a fallback key is used (typically when `Cargo.lock` or `Cargo.toml` is changed), but are quite significant. The only downside is an ~2 second overhead in the cache action from downloading `cargo-sweep`.

janhohenheim:
That's really impressive :D Good work!

benfrankel (approved):

PR #293, opened 2024-09-03
Fix `load_resource` stops proceeding if any of asset loading failed

akimakinai:
`Vec::pop`/`Vec::push` touches only the last element of `Vec`. So, if the asset there failed to load, `load_resource_assets` wouldn't process remaining asset any more. This PR changes it to use `VecDeque::push_back`/`VecDeque::pop_front` instead.


janhohenheim:
@cart looks like your code made it back in :D 
Could you add a comment explaining why we are using a `VecDeque` and not a `Vec`? lgtm otherwise.

benfrankel (approved):
Bug and fix seem correct to me from the code. Approved, but deferring to @janhohenheim's review before merging.

Thanks for fixing this :)

janhohenheim (approved):
Took the liberty of applying @benfrankel's suggestion ;)

Issue #300, opened 2024-09-09
cargo generate: crate name in doc comments is not replaced causing `cargo test` to fail on doc tests


lenhosseini:
After trying the template with cargo-generate, the crate name in the doc comments example in `src/audio.rs` is not replaced with the new crate name and therefore causes `cargo test` to fail. 

### src/audio.rs - Line 10

```rs
  use bevy_quickstart::audio::Music;
```

### src/audio.rs - Line 28

```rs
  use bevy_quickstart::audio::SoundEffect;
```

benfrankel:
Nice catch, thanks!

benfrankel:
Fixed in https://github.com/TheBevyFlock/bevy_quickstart/commit/6d316cc084cdd38e3d5e1202454e6a65d73bcf2f and https://github.com/TheBevyFlock/bevy_quickstart/commit/db5bfb16935efb91ddc24ed95a26eaecd4942261.

PR #301, opened 2024-09-09
Fix generated doctests failure

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_quickstart/issues/300.


benfrankel:
Hm okay this doesn't work. This will need to be updated on the `cargo-generate` branch instead.

Issue #288, opened 2024-08-28
Set `CARGO_GENERATE_PAT` using a maintainer's token


BD103:
This is targeted towards @alice-i-cecile or @mockersf, since they are owners of this org and have write permissions for this repo.

---

1. Go to <https://github.com/settings/tokens?type=beta> and click "Generate new token".
2. On the setup screen, set the name to "bevy_quickstart_cargo_generate" (or anything else that reminds you what it is).
3. Set the expiration date to "Custom..." and choose a date 1 year in the future. (Github won't let you set it farther than that!) In a year, you'll have to regenerate this token.
4. Set the description to "Used to automatically updated the `cargo-generate` branch in the `bevy_quickstart` repository." (Or anything else! You choose!)
5. Set the "Resource owner" as "TheBevyFlock", since that's where `bevy_quickstart` lives.
6. Under "Repository access", click "Only select repositories" and select "bevy_quickstart" from the dropdown.
7. If you did everything correctly, it should look like this:

![image](https://github.com/user-attachments/assets/72a50bcf-eb75-4be7-a61e-d7dbc5efe191)

8. Great! Next, scroll down to "Permissions" and click the "Repository permissions" box.
9. Set "Contents" to "Read and write" and "Workflows" to "Read and write". "Metadata" will automatically enable itself to be "Read-only", but the rest should remain "No access".
10. Scroll down to "Overview". If you did everything correctly, it should look like this:

![image](https://github.com/user-attachments/assets/c40df768-befc-4e6f-946c-86e14c8f5e83)

11. Click "Generate token" and copy the token displayed on screen.
12. Navigate to <https://github.com/TheBevyFlock/bevy_quickstart/settings/secrets/actions> and click "New repository secret".
13. Set the name to "CARGO_GENERATE_PAT" and paste your copied token into the "Secret" box.
14. If you did everything correctly, it should look like this:

![image](https://github.com/user-attachments/assets/c97a7c30-5388-41ae-bce2-30aeecbf064b)

15. Click "Add secret". Everything should be setup now!
16. To test that it now works, navigate to <https://github.com/TheBevyFlock/bevy_quickstart/actions/workflows/cargo-generate.yaml>, click "Run workflow", then click the new green "Run workflow" button.
17. The spawned workflow should either pass or log an error about merge conflicts. You're good to go!

## Regenerating the PAT

[...]

PR #305, opened 2024-09-16
Fuse `cargo-generate` with `main` branch

benfrankel:
Related to https://github.com/TheBevyFlock/bevy_quickstart/issues/288.

Testing:
- `cargo generate -ap path/to/bevy_quickstart` works
- `cargo run` in the generated repo works
- `cargo run` in the template itself works

Remaining tasks after this:

- [ ] Update `README.md` (remove manual clone instructions, simplify `cargo generate` commands)
- [ ] Remove "GitHub template" functionality
- [ ] Delete `cargo-generate.yaml` workflow
- [ ] Delete `cargo-generate` branch
- [ ] Add `ci-generate.yaml` workflow to make sure `cargo-generate` works and the generated repo is functional (this can be workshopped in https://github.com/TheBevyFlock/bevy_new first)

### DIFF
--- a/.github/workflows/release.yaml.template
+++ b/.github/workflows/release.yaml.template
@@ -1,0 +1,316 @@
+name: Release
+
+on:
+  # Trigger this workflow when a tag is pushed in the format `v1.2.3`.
+  push:
+    tags:
+      # Pattern syntax: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+      - "v[0-9]+.[0-9]+.[0-9]+*"
+  # Trigger this workflow manually via workflow dispatch.
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -37,7 +37,7 @@ impl Plugin for AppPlugin {
                 })
                 .set(WindowPlugin {
                     primary_window: Window {
-                        title: "bevy_quickstart".to_string(),
+                        title: "Bevy Quickstart".to_string(),
```
richchurcher, line 40: Just curious if this might be changing, with the repo moving to `bevy_new_2d`, will we be referring to it as `2D Template` or similar?

richchurcher (approved):
LGTM! Nice.

src/lib.rs
benfrankel: Ah yeah, we probably will. Can be a separate PR when the rename happens, though.

janhohenheim (approved):

PR #312, opened 2024-09-29
Switch from `trunk` to `bevy run web`

### LINKED ISSUE #297: Switch from `Trunk` to `bevy run web`
Blocked on https://github.com/TheBevyFlock/bevy_cli/issues/8

TimJentzsch:
# Objective

Closes #297.

This PR prepares the project to use the Bevy CLI, mainly the `bevy run` command.
It simplifies especially running for the web target, which previously required use of `trunk` (and a bunch of additional configuration).

# Solution

- Remove custom web assets. Use the defaults provided by Bevy CLI instead.
- Remove custom compilation profiles for the web build. The CLI will automatically use custom `web` and `web-release` profiles which use sensible defaults for web builds.
- Adjust the release CI workflow to use the Bevy CLI with `--bundle` instead of `trunk`.
- Update the README to use Bevy CLI instead of `trunk`.

Note that this introduces a "regression" to the user that they now have to specify `--no-default-features`, which was automatically passed to `trunk`.
I created <https://github.com/TheBevyFlock/bevy_cli/issues/250> so we can do that in Bevy CLI as well.

# Testing

1. Install the Bevy CLI:
    ```
    cargo install --git https://github.com/TheBevyFlock/bevy_cli --locked bevy_cli
    ```
2. Checkout this branch: `gh pr checkout 312` (with the GitHub CLI)
3. Run the app in the browser:
    ```
    bevy run --no-default-features web --open
    ```


TimJentzsch:
Current status:

- Release CI needs testing (for web), @benfrankel do you know an easy way to do that?
- Web sound is not working yet locally. Warning in console:
    ```
    An AudioContext was prevented from starting automatically. It must be created or resumed after a user gesture on the page.
    ```

benfrankel:
> Release CI needs testing (for web), @benfrankel do you know an easy way to do that?

The release workflow can be triggered manually, but not from a PR, I believe. There are a few options:

1. Merge this PR once it's ready and been reviewed, then test the release CI and make a fix PR if necessary.
2. Merge this PR in a temporary branch and run it there.
3. Run the release CI in a fork.

TimJentzsch:
The audio will be fixed by <https://github.com/TheBevyFlock/bevy_cli/pull/249>.

TimJentzsch:
@benfrankel This should be ready now.

For the CI I'm still uncertain if it already works as expected, but I'm fine with the options you proposed.
Option 1 is probably the easiest, but also with a bit of risk. But I guess it wouldn't break the already deployed version, so it's not that big of a deal.

Cargo.toml
```diff
@@ -61,28 +61,3 @@ opt-level = 3
...
-strip = "debuginfo"
-
-# Override some settings for native builds.
-[profile.release-native]
-# Default to release profile values.
-inherits = "release"
-# Optimize with performance in mind.
-opt-level = 3
-# Keep debug information in the binary.
-strip = "none"
```
benfrankel, line 88: Does `bevy build` include all of this already?

.vscode/tasks.json
```diff
@@ -25,25 +23,18 @@
             "label": "Run native release",
             "type": "cargo",
             "command": "run",
-            "args": [
-                "--release",
-                "--no-default-features"
-            ],
+            "args": ["--release", "--no-default-features"],
```
benfrankel, line 26: I'm curious why we suggest that users use the Bevy CLI for native release while the `.vscode/tasks.json` still uses `cargo run` directly. Is VS Code special-cased to interpret the output of `cargo` commands? Also, is there a benefit to `bevy run` over `cargo run` for native builds, or is it for uniformity?

.github/workflows/release.yaml
```diff
@@ -159,15 +159,15 @@ jobs:
       - name: Prepare output directories
         run: rm -rf tmp; mkdir -p tmp/binary '${{ env.app }}'
 
-      - name: Install cargo-binstall (Web)
+      - name: Install and run Bevy CLI (Web)
         if: ${{ matrix.platform == 'web' }}
-        uses: cargo-bins/cargo-binstall@v1.9.0
+        run: |
+          cargo install --git=https://github.com/TheBevyFlock/bevy_cli --locked bevy_cli
```
benfrankel, line 165: I suspect that this will slow down the workflow to some degree because it'll need to compile `bevy_cli`, whereas `cargo-bins/cargo-binstall` installs `cargo-binstall` from binary, which is then used to install `trunk` from binary. Maybe `cargo-binstall` could be used to install `bevy_cli` in the future (if not now? idk how `cargo-binstall` works).

.github/workflows/release.yaml
```diff
@@ -10,7 +10,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'Version number in the format `v1.2.3`'
+        description: "Version number in the format `v1.2.3`"
```
benfrankel, line 13: What's the reason for changing from single to double quotes? AFAICT the only difference is that double quotes allows escape sequences such as `\n`. Same question about the `run: "true"` line below.

benfrankel (commented):
I did a first pass on the diff, and I have a few questions.

.github/workflows/release.yaml
TimJentzsch: I'm sorry, that was my formatter when I saved the file.
I can revert it if you like

Cargo.toml
TimJentzsch: We don't set thin lto and the single codegen unit by default.

So the default `release` profile for Bevy is just the default that cargo provides.
For web builds we change the opt level to be `s` and inherit the `release` settings.
I plan to optimize the web profile further in the future.

But I can restore the native release settings!

web/index.html
```diff
@@ -1,38 +0,0 @@
-<!doctype html>
-<html lang="en">
-
-<head>
-    <meta charset="utf-8" />
-    <title>Bevy New 2D</title>
```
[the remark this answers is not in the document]
TimJentzsch: Right, I plan to show the binary name as title by default, but that's not implemented yet.

.vscode/tasks.json
TimJentzsch: Right now, the native build and run is just a wrapper around cargo.
Consistency makes sense though, I will adjust that

.github/workflows/release.yaml
TimJentzsch: From my testing it seems like binstall doesn't have a precompiled binary for Bevy CLI, so I thought we can save some time by not installing binstall.

I think it's because we don't have a GitHub release yet with the artifacts, but I'm also not sure how exactly binstall works.

But I assume we can eventually switch back to using cargo binstall!

.github/workflows/release.yaml
benfrankel: Ah okay, nbd either way then.

.github/workflows/release.yaml
benfrankel: Ok, this can be spun out as a followup issue once this PR is merged.

alec-deason:
For what it's worth I've been using these changes in a project and they've worked great for me.

web/style.css
```diff
@@ -1,56 +0,0 @@
...
-
-    100% {
-        transform: rotate(360deg);
-    }
-}
-
-#bevy {
-    /* Hide Bevy app before it loads */
-    height: 0;
-}
```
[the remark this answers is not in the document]
TimJentzsch: We are not using a canvas with a specific ID (`#bevy`), because it requires that ID to be set in the Bevy code and we wanted it to work with as little additional configuration as possible.
But turns out, if there is no `canvas` available, it will just be added automatically.

So basically the entire `canvas` is removed in the template, so it doesn't need to be hidden anymore

.vscode/tasks.json
TimJentzsch: Actually, the task type is set to `cargo`, so there might actually be some special handling happening. That needs to be investigated further

web/style.css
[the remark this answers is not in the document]
benfrankel: Okay, and I'll assume the mutation observer that waits for an added `canvas` also works, replacing the current one that waits for a change to `#bevy` with `height > 0`.

Cargo.toml
benfrankel: I think we can fully copy from the `Release Mode Optimizations` section of Bevy's [quick start guide](https://bevyengine.org/learn/quick-start/getting-started/setup/#compile-with-performance-optimizations) now that native release is the default `release` profile.

web/index.html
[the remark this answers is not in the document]
benfrankel: Okay, since Bevy CLI supports the user swapping in their own custom `index.html`, this can be considered non-blocking.

It's worth the tradeoff of making this template simpler (plus some users might not even care about web builds), and telling users "the default `index.html` is reasonable, but at some point you'll want to make a local copy to customize it" sounds fine to me.

web/index.html
```diff
@@ -1,38 +0,0 @@
...
-</head>
-
-<body>
-    <div id="game" class="center">
-        <div id="loading-screen" class="center">
-            <span class="spinner"></span>
-        </div>
-
-        <canvas id="bevy">
-            Javascript and canvas support is required
```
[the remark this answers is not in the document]
benfrankel: Since Bevy CLI's `index.html` already has a mutation observer waiting for the `canvas` to spawn in, maybe it can update the canvas to include this or similar text?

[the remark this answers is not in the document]
TimJentzsch: I don't think that would work, because then we already know that JavaScript is working.

I can add a `noscript` element with a similar message instead though.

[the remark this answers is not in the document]
benfrankel: Ah yeah, you're right. You can do the `noscript` element for a JavaScript support message and the observer for a `canvas` support message?

Cargo.toml
TimJentzsch: So these two options are actually the default for this profile, but right now when we customize any values we need to define the full profile again.

Adopted the quick start configuration as suggested

TimJentzsch:
Putting this back to draft while we resolve the remaining regressions

web/style.css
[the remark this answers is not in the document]
TimJentzsch: It doesn't add the `bevy` ID, but I think that's not used right now, is it?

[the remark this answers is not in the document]
benfrankel: Correct, that should be fine.

benfrankel (approved):
Dropping my approval here because I believe everything has been addressed in the PR itself as well as in `bevy_cli` at this point :)

.vscode/tasks.json
benfrankel: This can be investigated and potentially changed in a future PR, no need to block on this.

benfrankel: We use `bevy run` in VS Code tasks now :)

PR #318, opened 2024-10-23
Update to bevy 0.15

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/320.

We'll probably want to set up a tag / versioning system soon.

The `ChildBuild` workaround I chose is probably not ideal, but it works.

Confirmed that the game compiles, runs, and seems fine.


richchurcher (approved):
Also checked and working here! No idea what to say about the `ChildBuild` issue, but guess it doesn't block a merge?

bas-ie (approved):

bas-ie:
BTW I think the CI failure here is caused by `dynamic_linking` no longer working with `cargo test --doc`, have just been encountering this in another third-party crate.

alec-deason:
I think this error can be fixed by adding `export LD_LIBRARY_PATH=`rustc --print target-libdir` or the equivalent to be CI job before `cargo test` runs.

benfrankel:
I can reproduce the CI failure on my own machine by running `cargo test --doc`. Running the same command on `main` also fails, so the CI failure is not related to this PR.

PR #325, opened 2025-02-14
Load config files and other asset based resources by path

alec-deason:
I find it useful to have config files which I can load from assets but access in game as simple resources. In this change I've extended the existing asset tracking system with a `load_resource_from_path` method to support that use case. When the dev-native feature is active it will also automatically add a system to watch the asset and reload the associated resource when it changes.

I've also added a config.ron file which uses this system to configure the max_speed attribute of the duck. I'm loading the ron file using bevy_common_assets, which is a new dependency. If you don't want to bring in that dependency I could write a simple custom AssetLoader just for the config file or remove the config file entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.95"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
+
+[[package]]
 name = "approx"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -472,6 +478,19 @@ dependencies = [
  "encase",
  "serde",
  "wgpu-types",
+]
+
+[[package]]
+name = "bevy_common_assets"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3521990269672c442f2bf0fbed0fce9db719e3dd136dd4012a97809464a4389d"
+dependencies = [
+ "anyhow",
+ "bevy",
+ "ron",
+ "serde",
+ "thiserror",
 ]
 
 [[package]]
@@ -885,8 +904,10 @@ name = "bevy_new_2d"
 version = "0.1.0"
 dependencies = [
  "bevy",
+ "bevy_common_assets",
  "log",
  "rand",
+ "serde",
  "tracing",
 ]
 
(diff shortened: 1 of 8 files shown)


benfrankel:
One of the core design goals of this template is to only depend on `bevy` itself, meaning no 3rd-party dependencies like `bevy_common_assets`. Also, this feature might be considered out of scope for this template, but that's more up for debate.

alec-deason:
That's a super reasonable design principle. I've removed the dependency on bevy_common_assets and the PR now only adds dependencies on crates bevy already uses.

I can absolutely understand deciding this is out of scope. But I do think that it's valuable to provide an easy and extensible way for new users to load this kind of data in a manner that is compatible with web builds and doesn't have the extra burden of needing to retrieve the asset from the assets store in every system. It's something that I personally want in every project which is why I think it's worth having in the template.

janhohenheim:
@alice-i-cecile could I get your thoughts on this from a boilerplate-to-feature perspective?

janhohenheim:
You mean as part of the Bevy examples, or as something else?

I myself am a bit torn, but I think the claim that this is something nearly everyone can profit from is true.
Let's not forget that this template in general is not shying away from opinionated patterns if it's something that most people new to Bevy will want, but may not know how to implement. So, for me personally, the opinionatedness is not much of a concern, 
I'm mostly thinking about how much this will make the experience of looking at the template as a beginner more complicated vs how useful it is to said beginner a week later.

alice-i-cecile:
> You mean as part of the Bevy examples, or as something else?

Yeah, as part of the Bevy examples. I've been chewing on "how do we communicate useful patterns without blessing them" to users, and this feels like it fits in there.

I see your point about this template being more opinionated, but I don't have strong feelings there. My personal template preferences are more minimal though: I basically want to never *delete* code from it because I don't like the pattern chosen.

janhohenheim:
That's fair. I'm tending towards "this is boilerplate every user would want and probably implement the same way, so let's make it easy for them". But I'm also fine with adding it to the official examples and then just linking it.
@benfrankel @MiniaczQ what's your opinion?

MiniaczQ:
Bit much boilerplate for game template

alec-deason:
> Yeah, as part of the Bevy examples. I've been chewing on "how do we communicate useful patterns without blessing them" to users, and this feels like it fits in there.

I think I've come around on the question of whether this belongs in the template, I agree it doesn't. I'd be happy to write up a more copy-and-pastable version for the bevy examples though. What section would you want it in, assets?

janhohenheim:
Yeah, `Assets` are a good choice :)
Closing this PR for now. Feel free to add a link to the example once it exists!

PR #344, opened 2025-03-29
Prefer `with_child` over `with_children`

benfrankel:
Also renamed the closure argument in `with_children` from `children` to `parent`, because that's the most common name used in `bevy` itself (there were 0 instances of `children`; the second-most common name was `builder`).

### DIFF
--- a/src/screens/loading.rs
+++ b/src/screens/loading.rs
@@ -18,8 +18,8 @@ fn spawn_loading_screen(mut commands: Commands) {
     commands
         .ui_root()
         .insert(StateScoped(Screen::Loading))
-        .with_children(|children| {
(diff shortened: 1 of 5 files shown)


src/screens/loading.rs
```diff
@@ -18,8 +18,8 @@ fn spawn_loading_screen(mut commands: Commands) {
     commands
         .ui_root()
         .insert(StateScoped(Screen::Loading))
-        .with_children(|children| {
-            children.label("Loading...").insert(Node {
+        .with_children(|parent| {
```
janhohenheim, line 21: Opened https://github.com/TheBevyFlock/bevy_new_2d/issues/345

janhohenheim (approved):

PR #346, opened 2025-03-30
Prefer `children!` macro

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/345.


janhohenheim (approved):
thanks for showing me how to use it, haha

PR #348, opened 2025-03-31
Improve CI usage awareness

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/212.

This adds some new flags to the release workflow:
- `use_github_cache` to toggle caching the release workflow, which is now _off_ by default because it unfortunately doesn't do anything unless the user releases multiple times _to the same tag_.
- `build_for_windows` etc. convenience toggles for individual platforms.

These changes can be tested (and fixed if necessary) after this PR is merged.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -68,7 +68,7 @@ env:
   # https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage
   git_lfs: false
 
-  # This only helps with consecutive releases to the same tag (and takes up cache storage space).
+  # Enabling this only helps with consecutive releases to the same tag (and takes up cache storage space).
   # See: https://github.com/orgs/community/discussions/27059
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -62,6 +68,10 @@ env:
   # https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage
   git_lfs: false
 
+  # This only helps with consecutive releases to the same tag (and takes up cache storage space).
```
janhohenheim, line 71:
```suggestion
  # Enabling this setting only helps with consecutive releases to the same tag (and takes up cache storage space).
```

Otherwise it's not clear whether the text refers to the `true` or false `variant`
Same for the other instances if this comment

janhohenheim (changes requested):
Love the new platform toggles!

janhohenheim (approved):

Issue #333, opened 2025-03-26
Support Jetbrains IDEs better


benfrankel:
Currently we have `.vscode/tasks.json` for VS Code, but nothing for Jetbrains IDEs like RustRover. If possible, we should add equivalent configuration, or at least explain in the `README` that it's not provided and users should configure the correct commands themselves.

See https://github.com/TheBevyFlock/bevy_new_2d/issues/327 for an example of a user issue due to the lack of support.

Sapein:
I've done some research and went ahead and opened a PR for this, but this is definitely possible. It's not necessarily as nice as the way it works in vscode and there are a bunch of other options, but at the very least the PR I've provided is a very minimal implementation of this. 

There are a few questions to resolve such as naming, making sure they're configured right, and so on. The biggest question I do have, which the PR doesn't resolve, is if we should use JetBrains's default .gitignore file for that folder, if we should have one at all, or not. Although, it looks like the JetBrains `.gitignore` file isn't generated if there's a `.idea` folder already in the directory, which might be something to keep in mind.

benfrankel:
There's some followup work to further improve the Jetbrains support (discussion in https://github.com/TheBevyFlock/bevy_new_2d/pull/360), but the basic support necessary to close this issue has been merged.

PR #360, opened 2025-04-19
Add in RustRover/Jetbrains Run Configurations

Sapein:
As seen in #333 there are some potential issues with running/supporting RustRover and other Jetbrains IDEs in that the default options that are provided are not necessarily what should actually be used. 

In Visual Studio Code, this is currently solved through the `.vscode/tasks.json` file. The configurations provided here are based off of what was in that file, and should be mostly equivalent. 

This adds in four shell command Build Configurations, using the bevy shell command. They are put under a Group called "Build", which is done to approximate how the vscode tasks work.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/.idea/runConfigurations/Run_Native_Release.xml
+++ b/.idea/runConfigurations/Run_Native_Release.xml
@@ -1,0 +1,17 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Run Native Release" type="ShConfigurationType" folderName="Build">
+    <option name="SCRIPT_TEXT" value="bevy run --release" />
+    <option name="INDEPENDENT_SCRIPT_PATH" value="true" />
+    <option name="SCRIPT_PATH" value="" />
+    <option name="SCRIPT_OPTIONS" value="" />
+    <option name="INDEPENDENT_SCRIPT_WORKING_DIRECTORY" value="true" />
+    <option name="SCRIPT_WORKING_DIRECTORY" value="$PROJECT_DIR$" />
+    <option name="INDEPENDENT_INTERPRETER_PATH" value="true" />
+    <option name="INTERPRETER_PATH" value="/bin/bash" />
+    <option name="INTERPRETER_OPTIONS" value="" />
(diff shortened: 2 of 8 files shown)


benfrankel:
Thanks for working on this. I don't use a Jetbrains IDE for Rust, so I won't be able to test this myself.

Sapein:
It's my pleasure! I figured that since I almost exclusively use RustRover for my Rust Development (especially for Bevy) I would probably be rather well suited to work on this. Everything *seems* to work more or less. 

I've tried to keep it as close to the current Visual Studio Code setup for the time being, and there are some questions that need to be answered, but I can confirm that this does work. 

I'm not super huge on having them all under a Build Folder as, to me, that feels a bit extraneous, but that's more-so because I don't really use Build Configuration Folders myself. 

The other main question is about if we should include the default `.gitignore` file provided by JetBrains when it generates the `.idea` folder, as by including this, the IDE will *NOT* generate that file. 

I do have some opinions on using Shell Commands, but ultimately it seems that the desired way to use bevy new 2d is through using bevy_cli. So it's what I went with.

janhohenheim:
Any chance we can also support the `plugin` snippet?

Sapein:
I've gone ahead and removed the Draft status from this, as this is the minimum we would need to have similar support on RustRover that we have on Visual Studio Code.

I've gone ahead and added in the autogenerated .gitignore that JetBrains's IDEs generate in the `.idea` folder when you first open a project, since that file won't be generated with this change, as the `.idea` folder exists. This file shouldn't really change, as it doesn't seem to ever be regenerated at this time. 

There was also some discussion on Discord about whether or not to use the Cargo Build Configurations which are included in RustRover by default, as it does provide slightly nicer UX. However, web builds would be a bit more involved if we use that instead due to the other commands that would need to be done. We can definitely go down that route in the future, however it would likely take some time and some investigation to get things dialed in (particularly for web). 

The only issue that I've found, is that you can't use the debugger in Rust Rover with the Terminal Commands. So we might want to include at least one Cargo Configuration that can be used to debug.

janhohenheim (approved):

Sapein:
> Any chance we can also support the `plugin` snippet?

`plugin` snippet?

Sapein:
All right so I looked into the snippet functionality, and there is such functionality in RustRover, and other JetBrains IDEs, they are called Live Templates. The issue is that, for some reason, they are *global* and not local to the project. 

With this, the only way we can really 'export' the templates is for me to port them to RustRover in my settings, then create an export of my settings with *only* live templates, and then anyone can import that file to get the live templates. 

There is a plugin that would, theoretically, allow this to work but it hasn't been updated to work with the latest RustRover at this time, and I'm not sure if saying "You have to install a third party plugin" for RustRover is the best idea.

benfrankel (approved):

PR #363, opened 2025-04-20
Tidy up `lib.rs` a little bit

benfrankel:
I couldn't reproduce the behavior the `IsDefaultUiCamera` comment was warning about, so I removed the comment and the component.

### DIFF
--- a/src/lib.rs.template
+++ b/src/lib.rs.template
@@ -16,7 +16,7 @@ pub struct AppPlugin;
 
 impl Plugin for AppPlugin {
     fn build(&self, app: &mut App) {
-        // Order new `AppStep` variants by adding them here:
+        // Order new `AppSet` variants by adding them here:
         app.configure_sets(
             Update,
             (AppSet::TickTimers, AppSet::RecordInput, AppSet::Update).chain(),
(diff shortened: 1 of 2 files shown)


src/lib.rs.template
```diff
@@ -80,15 +78,5 @@ enum AppSet {
...
-    commands.spawn((
-        Name::new("Camera"),
-        Camera2d,
-        // Render all UI to this camera.
-        // Not strictly necessary since we only use one camera,
-        // but if we don't use this component, our UI will disappear as soon
-        // as we add another camera. This includes indirect ways of adding cameras like using
-        // [ui node outlines](https://bevyengine.org/news/bevy-0-14/#ui-node-outline-gizmos)
-        // for debugging. So it's good to have this here for future-proofing.
-        IsDefaultUiCamera,
```
janhohenheim, line 92: I think omitting this is a footgun

janhohenheim (approved):

src/lib.rs.template
benfrankel: I see, it must have changed in 0.16 then.

PR #364, opened 2025-04-20
Add hex values to docs

janhohenheim:
This way, VS Code and I *believe* RustRover show some lovely previews:
image


janhohenheim:
@Sapein mind checking this on RustRover real quick?

Sapein:
> @Sapein mind checking this on RustRover real quick?

This actually doesn't seem to occur in Rust Rover, although I'll be honest I would've sworn it did work.

janhohenheim:
@Sapein doing so looks like this in VS Code:
image
And when hovering over:
image

Sapein:
@janhohenheim Interesting. This is how it looks in RustRover

Just looking at it
color1

On Hover
color2

benfrankel:
The previews are definitely nice to have, but I'm not sure they're worth the cost to users who would have to keep the comments and code in sync whenever they change the colors.

janhohenheim:
@benfrankel wouldn't that be the choice of the users, then?

benfrankel:
Well, they'd have to either delete the comments or keep them in sync.

benfrankel:
Ideally we'd be pulling these colors from a config file where they could be specified as hex codes for IDE support, instead of hard-coding them as `const`s in the code. See https://github.com/TheBevyFlock/bevy_new_2d/issues/31.

janhohenheim:
@benfrankel fully agreed that that would be best

janhohenheim:
I would still include this as people would realistically copy-paste both values from a color picker and not fiddle with the constants in code directly, but I may be outnumbered here.
Should we close this?

benfrankel:
I mean I'm a little bit on the fence, there is value in having hex codes here, I just don't love the duplication. It's not a big deal either way, though.

PR #365, opened 2025-04-21
Add `bevy_lint` support

BD103:
This PR adds [`bevy_lint`](https://thebevyflock.github.io/bevy_cli/bevy_lint/) support to the template (the linter is a sister project of the Bevy CLI, both are in the same [repository](https://github.com/TheBevyFlock/bevy_cli/tree/main/bevy_lint)).

This PR is currently a work-in-progress, but I'll mark it as ready-for-review when it's done! :)


.github/workflows/ci.yaml
```diff
@@ -103,3 +103,40 @@ jobs:
 
       - name: Check documentation
         run: cargo doc --locked --workspace --all-features --document-private-items --no-deps
+
+  bevy-lint:
+    name: Bevy linter
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      LINTER_TOOLCHAIN: nightly-2025-04-03
```
janhohenheim, line 112: This is specific to the linter version installed, right? If so, can we add a comment? Maybe a link to that fancy md table, noting that needs to be updated for future versions of Bevy?
Side note, maybe that could be encoded into an own GitHub workflow at some point :)

BD103: Actually, let me just make this an action in the `bevy_cli` repository. That should work better!

BD103:
I created https://github.com/TheBevyFlock/bevy_cli/pull/380, but this PR is blocked until that is merged :)

BD103:
This should be good to go now! This PR now uses the official action for installing `bevy_lint`, so that should hopefully mean less maintenance work for you all :)

Cargo.toml
```diff
@@ -54,6 +54,10 @@ default_features = false
 features = ["dev"]
 
 
+[lints.rust]
+# Mark `bevy_lint` as a valid `cfg`, as it is set when the Bevy linter is run.
```
benfrankel, line 58:
```suggestion
# Mark `bevy_lint` as a valid `cfg`, as it is set when the Bevy linter runs.
```

Cargo.toml.template
```diff
@@ -48,6 +48,10 @@ default_features = false
 features = ["dev"]
 
 
+[lints.rust]
+# Mark `bevy_lint` as a valid `cfg`, as it is set when the Bevy linter is run.
```
benfrankel, line 52:
```suggestion
# Mark `bevy_lint` as a valid `cfg`, as it is set when the Bevy linter runs.
```

.github/workflows/ci.yaml
```diff
@@ -132,3 +132,29 @@ jobs:
 
       - name: Check documentation
         run: cargo doc --locked --workspace --all-features --document-private-items --no-deps
+
+  bevy-lint:
```
benfrankel, line 136:
```suggestion
  # Run Bevy lints.
  bevy-lint:
```

.github/workflows/ci.yaml
```diff
@@ -132,3 +132,29 @@ jobs:
...
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Install Bevy linter
+        uses: TheBevyFlock/bevy_cli/bevy_lint@lint-v0.3.0
+
+      - name: Run Bevy linter
+        run: bevy_lint --locked --workspace --all-targets --all-features
+        env:
+          RUSTFLAGS: --deny warnings
```
benfrankel, line 160:
```suggestion
        env:
          RUSTFLAGS: --deny warnings
        run: bevy_lint --locked --workspace --all-targets --all-features
```

benfrankel (changes requested):
Small nits, looks good otherwise.

janhohenheim (approved):

.github/workflows/ci.yaml
benfrankel: Or "Run Bevy linter.", fine either way, I wrote it this way to parallel the "Run clippy lints." comment above.

benfrankel:
I see the action installs `bevy_lint` with `cargo install`, which builds it from source. It seems to only take ~1m in CI and the job runs in parallel with the other CI jobs so this is not a big deal for now. We do want to be friendly to private repositories, so `cargo-binstall` for `bevy_lint` in the future would be nice.

BD103:
> I see the action installs `bevy_lint` with `cargo install`, which builds it from source. It seems to only take ~1m in CI and the job runs in parallel with the other CI jobs so this is not a big deal for now. We do want to be friendly to private repositories, so `cargo-binstall` for `bevy_lint` in the future would be nice.

Yup! Pre-built binaries are difficult with the linter due to all its other prerequisites, but definitely possible! If you're concerned about CI minutes, I'd focus on getting the Bevy CLI pre-built first :)

benfrankel (approved):

PR #366, opened 2025-04-22
Improve UI

benfrankel:
Summary of changes:
- Code:
  - Use `fn() -> impl Bundle` when possible (replacing all the traits in `widgets` module)
  - Rename `widgets` -> `widget`
- Buttons:
  - Trigger on _release_, not on press, since we can do that now
  - Add `BorderRadius::MAX`
  - Tweak background colors
- Credits screen:
  - Use grid nodes for movie credits-style layout
  - Change header text color and remove background color

I haven't tested how these changes look in the itch.io embed yet, the credits screen might not fit in there because everything is specified in `Px` instead of `Vw` etc.


src/screens/credits.rs
```diff
@@ -15,24 +15,74 @@ pub(super) fn plugin(app: &mut App) {
...
+}
 
-            parent.header("Assets");
-            parent.label("Bevy logo - All rights reserved by the Bevy Foundation. Permission granted for splash screen use when unmodified.");
-            parent.label("Ducky sprite - CC0 by Caz Creates Games");
-            parent.label("Button SFX - CC0 by Jaszunio15");
-            parent.label("Music - CC BY 3.0 by Kevin MacLeod");
+fn assets() -> impl Bundle {
+    grid(vec![
+        "Ducky sprite",
```
janhohenheim, line 46: Could we make this a vector of tuples?

src/theme/widget.rs
```diff
@@ -0,0 +1,71 @@
+//! Helper functions for creating common widgets.
+
+use bevy::{prelude::*, ui::Val::*};
+
+use crate::theme::{interaction::InteractionPalette, palette::*};
+
+/// Spawn a root node that covers the full screen
```
janhohenheim, line 7: `Spawn` may be the wrong verb here

janhohenheim:
Didn't test anything, but the code changes are lovely! :)

src/theme/widget.rs
benfrankel: Ah true.

src/screens/credits.rs
benfrankel: Done.

janhohenheim (approved):

benfrankel:
I checked that the credits screen fits in a window that's the same size as the itch.io embed.

PR #369, opened 2025-04-22
Embed observer within button widget

benfrankel:
Thanks to [DreamerTalin](https://discord.com/channels/691052431525675048/885021580353237032/1359605414013960253) and [Umatriz](https://discord.com/channels/691052431525675048/885021580353237032/1364100942793805884) for figuring out the `impl SpawnableList<ChildOf> for SpawnObserver` trick.

### DIFF
--- a/src/theme/widget.rs
+++ b/src/theme/widget.rs
@@ -82,13 +82,14 @@ pub fn button<E: Event, B: Bundle, M, I: IntoObserverSystem<E, B, M>>(
 }
 
 /// A [`SpawnableList`] that spawns an [`Observer`] as a child entity.
-struct SpawnObserver(pub Observer);
+struct SpawnObserver(Observer);
 
 impl SpawnableList<ChildOf> for SpawnObserver {
     fn spawn(self, world: &mut World, entity: Entity) {
         world.spawn(self.0.with_entity(entity));
     }
 
+    // Size hint is not important for this simple use case, so return 0.
     fn size_hint(&self) -> usize {
         0
     }
--- a/src/theme/widget.rs
+++ b/src/theme/widget.rs
@@ -3,7 +3,7 @@
 use std::borrow::Cow;
 
 use bevy::{
-    ecs::{spawn::SpawnableList, system::IntoObserverSystem},
+    ecs::{relationship::RelatedSpawner, spawn::SpawnWith, system::IntoObserverSystem},
     prelude::*,
     ui::Val::*,
 };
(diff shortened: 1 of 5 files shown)


src/theme/widget.rs
```diff
@@ -44,31 +69,27 @@ pub fn button(text: impl Into<String>) -> impl Bundle {
...
-pub fn header(text: impl Into<String>) -> impl Bundle {
-    (
-        Name::new("Header"),
-        Text(text.into()),
-        TextFont::from_font_size(40.0),
-        TextColor(HEADER_TEXT),
-    )
-}
+/// A [`SpawnableList`] that spawns an [`Observer`] as a child entity.
+struct SpawnObserver(pub Observer);
```
janhohenheim, line 85: The struct is not pub, but the Observer inside is. Is that correct?

src/theme/widget.rs
```diff
@@ -44,31 +69,27 @@ pub fn button(text: impl Into<String>) -> impl Bundle {
...
-        TextColor(LABEL_TEXT),
-    )
+impl SpawnableList<ChildOf> for SpawnObserver {
+    fn spawn(self, world: &mut World, entity: Entity) {
+        world.spawn(self.0.with_entity(entity));
+    }
+
+    fn size_hint(&self) -> usize {
+        0
+    }
```
janhohenheim, line 94: Can we annotate that this is meaningless boilerplate?

src/theme/widget.rs
benfrankel: Right, I removed the `pub` from the struct and forgot to remove it from its field.

src/theme/widget.rs
benfrankel: Sure

janhohenheim (approved):

ChristopherBiscardi:
From the perspective of someone who will have to present this to people to use, building brand-new abstractions that don't exist anywhere else in the ecosystem is a very strange speedbump to deal with. There's no good place to point people to, to teach them why `SpawnObserver` only seems to exist in the template and not in their other Bevy applications. I kind of *have* to cover the extra abstractions if I talk about the template because its a departure from the way other Bevy applications work.

Pedagogically, given that in this PR, you would *already* need to understand `Children::spawn`/`Spawn` style APIs (and more) to understand the `SpawnObserver` in `button`, going directly to using the existing APIs seems like it would remove the need for the abstraction entirely.

```rust
    commands.spawn((
        Node::default(),
        Children::spawn((
            SpawnWith(
                |parent: &mut RelatedSpawner<ChildOf>| {
                    parent
                        .spawn(button("Forward"))
                        .observe(forward_observer);
                },
            ),
            Spawn(button("Back")),
        )),
    ));
```

janhohenheim (changes requested):
Let's use the SpawnWith variant then :)

benfrankel:
Done. Basically I wanted to encapsulate the fact that the button entity is actually being spawned within a `SpawnWith` to give access to `.observe(...)` inside the `fn button()`, which is why I had to add a new outer entity. The buttons still look and function correctly this way.

janhohenheim (approved):
Lovely!

PR #372, opened 2025-04-25
Some style changes

benfrankel:
Please let me know if you don't like any of these changes, because they're subjective.

Summary of changes:
- Simplify comments in `audio.rs` and add bundle functions for sfx/music
- Split a `register_type::<(...)>` (see [cart's reasoning on Discord](https://discord.com/channels/691052431525675048/1002362493634629796/1351279139872571462))
- Use `Single` where possible (only one place)
- Use the `ChildSpawner` type alias instead of `RelatedSpawner<ChildOf>`
- Remove `I: Sync` trait bound by calling `IntoObserverSystem::into_system(action)` outside of the closure

I also looked for any cases where we could use Bevy's new error handling with `-> Result`, but we don't seem to have any good use cases for it.

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -1,37 +1,25 @@
 use bevy::prelude::*;
 
-/// An organizational marker component that should be added to a spawned [`AudioPlayer`] if it is in the
-/// general "music" category (ex: global background music, soundtrack, etc).
+/// An organizational marker component that should be added to a spawned [`AudioPlayer`] if it's in the
+/// general "music" category (e.g. global background music, soundtrack).
 ///
-/// This can then be used to query for and operate on sounds in that category. For example:
-///
-/// ```
-/// use bevy::{audio::Volume, prelude::*};
-/// use bevy_new_2d::audio::Music;
-///
-/// fn set_music_volume(mut sink_query: Query<&mut AudioSink, With<Music>>) {
-///     for mut sink in &mut sink_query {
-///         sink.set_volume(Volume::Linear(0.5));
-///     }
-/// }
-/// ```
+/// This can then be used to query for and operate on sounds in that category.
 #[derive(Component, Default)]
 pub struct Music;
 
-/// An organizational marker component that should be added to a spawned [`AudioPlayer`] if it is in the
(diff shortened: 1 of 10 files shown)


src/screens/credits.rs
```diff
@@ -82,34 +82,26 @@ fn enter_title_screen(_: Trigger<Pointer<Click>>, mut next_screen: ResMut<NextSt
...
-        commands
-            .spawn((
-                AudioPlayer(music.music.clone()),
-                PlaybackSettings::LOOP,
-                Music,
-            ))
-            .id(),
-    );
+fn start_credits_music(mut commands: Commands, mut credits_music: ResMut<CreditsMusic>) {
+    credits_music.entity = Some(commands.spawn(music(credits_music.handle.clone())).id());
```
janhohenheim, line 100: All of this on one line is quite hard to read. Could you separate this into variables?

src/audio.rs
```diff
@@ -1,37 +1,25 @@
...
-/// An organizational marker component that should be added to a spawned [`AudioPlayer`] if it is in the
-/// general "music" category (ex: global background music, soundtrack, etc).
+/// An organizational marker component that should be added to a spawned [`AudioPlayer`] if it's in the
+/// general "music" category (e.g. global background music, soundtrack).
 ///
-/// This can then be used to query for and operate on sounds in that category. For example:
-///
-/// ```
-/// use bevy::{audio::Volume, prelude::*};
-/// use bevy_new_2d::audio::Music;
```
janhohenheim, line 10: Finally bit the bullet on visibility, eh? :D

janhohenheim, line 10: Finally bit the bullet on visibility, eh? :D

janhohenheim (changes requested):

janhohenheim:
Changed look quite clean overall to me, don't think any of this is controversial

src/screens/credits.rs
benfrankel: Ok, I'll split out the handle.

janhohenheim (approved):

PR #373, opened 2025-04-25
Move loading sync point to before gameplay

janhohenheim:
- This way, assets can load during the splash screen and title menu.
- Restarting the app will lead to visiting the loading state multiple times, but the UI will only show when actually loading anything
- We didn't preload the splash image before and are still not doing that because I'd prefer to [just use embedded_asset](https://github.com/bevyengine/bevy/issues/14246) once it works on all platforms

### DIFF
--- a/src/screens/loading.rs
+++ b/src/screens/loading.rs
@@ -10,7 +10,7 @@ pub(super) fn plugin(app: &mut App) {
 
     app.add_systems(
         Update,
-        continue_to_title_screen.run_if(in_state(Screen::Loading).and(all_assets_loaded)),
+        continue_to_gameplay_screen.run_if(in_state(Screen::Loading).and(all_assets_loaded)),
     );
 }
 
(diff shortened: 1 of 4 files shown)


src/screens/loading.rs
```diff
@@ -6,11 +6,14 @@ use bevy::prelude::*;
 use crate::{asset_tracking::ResourceHandles, screens::Screen, theme::prelude::*};
 
 pub(super) fn plugin(app: &mut App) {
-    app.add_systems(OnEnter(Screen::Loading), spawn_loading_screen);
+    app.add_systems(
+        OnEnter(Screen::Loading),
+        spawn_loading_screen.run_if(not(all_assets_loaded)),
```
benfrankel, line 11:
```suggestion
        spawn_loading_screen.run_if(not(all_assets_loaded)),
```
I'd rather move this check to `splash::continue_to_title_screen` (renamed ofc). It's better to skip the loading screen entirely if there's nothing to load, instead of flashing the app's clear color for 1 frame.

benfrankel (changes requested):
Looks good other than this one comment.

src/screens/loading.rs
janhohenheim: Hmm? The splash does not really have anything to do with the loading screen now. This check here is for when the player presses `Esc` to restart the game, which bypasses the splash entirely

benfrankel: My bad, replace `splash::continue_to_title_screen` with `title::enter_loading_screen`. 

(also off-topic, I noticed we're naming the systems `continue_to` vs. `enter` when it may be reasonable to name them all `enter`).

janhohenheim: Done: https://github.com/TheBevyFlock/bevy_new_2d/pull/373/commits/b11db8a317ce1d1a9919ecb4c8c2a16e4b597e3b

benfrankel (approved):

PR #374, opened 2025-04-25
Implement a settings menu

### LINKED ISSUE #213: The template should come with a settings menu
There should be a settings menu, that at the very least implements audio volume settings (or two toggles for muting music and game sound independently).

janhohenheim:
image

Fixes #213 by adding a menu for volume control.

I know this was deemed out of scope a long time ago, but I want to resurrect that discussion, mainly for the sake of accessibility. I'd like to live in a world where Bevy jam games come with accessibility options (e.g. mouse acceleration, FOV, etc.) out of the box, and IMO that is only plausible if we give devs and users a settings menu for that.
The historical answer to this was to add it to the Bevy examples for users to look up, but
- a) we all know that a jam is on a tight deadline, so I don't expect users to spend extra time on this.
- b) Bevy should imo be as accessible out of the box as possible, and I want to extend that to this template.
- c) a settings menu is something that pretty much every single game in existence will profit from, even if only to allow audio control for the players. Many browsers don't even allow lowering the volume in a tab without muting it completely.

The other worry was that this would bloat up the template. To address this:

- d) the boilerplate for the settings screen is largely contained in the settings module, so it's supremely easy to rip it out if you don't want it.
- e) I intentionally wrote the code so that no new abstraction is introduced. Users don't need to know about this, they are free to entirely ignore it.

## Future work

- If this is merged, a future PR can make the menu accessible in-game through pausing. I left that task intentionally to a followup, as its implementation is largely orthogonal to this.
- We could also add new options to this menu, like screen resolution, MSAA samples, etc.


src/audio.rs
```diff
@@ -1,5 +1,19 @@
+use bevy::audio::Volume;
 use bevy::prelude::*;
 
+pub(super) fn plugin(app: &mut App) {
+    app.init_resource::<GlobalMusicVolumeScale>();
+    app.register_type::<GlobalMusicVolumeScale>();
+    app.add_systems(
+        PostUpdate,
```
janhohenheim, line 8: Running in `PostUpdate` to guarantee this frame has the correct volume

src/audio.rs
```diff
@@ -33,5 +48,33 @@ pub struct Music;
...
-#[derive(Component, Default)]
+#[derive(Component, Debug, Default)]
 pub struct SoundEffect;
+
+#[derive(Resource, Debug, Reflect, Deref, DerefMut, Default)]
+#[reflect(Resource)]
+pub struct GlobalMusicVolumeScale(pub Volume);
+#[derive(Component, Debug, Reflect, Deref, DerefMut, Default)]
+#[reflect(Component)]
+pub struct UnscaledVolume(pub Volume);
```
janhohenheim, line 59: Has no `Music` in the name because I'd reuse this for SFX

src/screens/settings.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+struct PercentageSettings<E, B1, B2, M1, M2, I1, I2, C>
+where
+    E: Event,
+    B1: Bundle,
+    B2: Bundle,
+    I1: IntoObserverSystem<E, B1, M1> + Sync,
+    I2: IntoObserverSystem<E, B2, M2> + Sync,
+    C: Component,
```
janhohenheim, line 51: I hate this

src/screens/settings.rs
```diff
@@ -0,0 +1,138 @@
...
+        Update,
+        update_music_volume_label.run_if(in_state(Screen::Settings)),
+    );
+}
+
+#[derive(Component)]
+struct MusicVolumeLabel;
+
+fn spawn_settings_screen(mut commands: Commands) {
+    let music_settings = PercentageSettings {
```
janhohenheim, line 26: I'm using a struct here and not passing it as params because I think passing 4 weird params looks unreadable. Especially if you happen to forget whether `on_minus` or `on_plus` goes first.

src/screens/settings.rs
```diff
@@ -0,0 +1,138 @@
...
+        name: "Music Volume".into(),
+        value_marker: MusicVolumeLabel,
+        on_minus: lower_music_volume,
+        on_plus: raise_music_volume,
+        _marker: PhantomData,
```
janhohenheim, line 31: I don't like that the `PhantomData` is constructed by hand, but `..default()` wouldn't make sense, as there is no sane default for these params, and `::new` would eliminate the positional param names

src/theme/widget.rs
```diff
@@ -88,3 +88,45 @@ where
         })),
     )
 }
+
+/// A simple button with text and an action defined as an [`Observer`].
+pub fn button_small<E, B, M, I>(text: impl Into<String>, action: I) -> impl Bundle
```
janhohenheim, line 93: Largely copy-pasted from `button`, dunno if that's considered bad form in UI code or not. I imagine this is better than introducing de-duplication functions, as you may want to fine-tune all components in this hierarchy.

will-hart: To remove duplication I renamed the `button` from this PR into `button_base` which takes a size and font size argument. Then I have:

```rust
pub fn button<E, B, M, I>(text: impl Into<String>, action: I) -> impl Bundle
where
    E: Event,
    B: Bundle,
    I: IntoObserverSystem<E, B, M> + Sync,
{
    button_base(text, action, Vec2::new(200.0, 65.0), FontSizes::HEADER)
}

pub fn button_small<E, B, M, I>(text: impl Into<String>, action: I) -> impl Bundle
where
    E: Event,
    B: Bundle,
    I: IntoObserverSystem<E, B, M> + Sync,
{
    button_base(text, action, Vec2::new(180.0, 35.0), FontSizes::UI)
}

```

For my needs at least this is fine, font size and size are the main things that change for these kinds of "standard" buttons.

janhohenheim: That's neat, I'll copy that design :)

benfrankel:
Have you seen [`GlobalVolume`](https://docs.rs/bevy/latest/bevy/audio/struct.GlobalVolume.html)?

janhohenheim:
@benfrankel I did not! TIL!
But that cannot distinguish between SFX and music, right?
Ofc, we could just ignore that distinction for the template.
Or add it in a different PR.

benfrankel:
IMO we should ignore the distinction for simplicity. We'll want a master volume slider one way or another, so doing only that for now seems good for a settings menu MVP.

janhohenheim:
Addressed your feedback. I also changed the volume of the files so that 100% is tolerable now. Dunno why the files got a bit bigger, but I will not fuss about a KiB.

src/theme/widget.rs
```diff
@@ -47,8 +47,55 @@ pub fn label(text: impl Into<String>) -> impl Bundle {
...
+            height: Px(30.0),
+            align_items: AlignItems::Center,
+            justify_content: JustifyContent::Center,
+            ..default()
+        },
+    )
+}
+
+/// A simple button with text and an action defined as an [`Observer`]. The button's components will be overridden by the components in `button_bundle`.
+pub fn button_base<E, B, M, I>(
```
benfrankel, line 94: I'd rather explicitly make this a non-pub helper function. Also, the `.insert(button_bundle)` overwriting trick is not needed anymore since you removed `Node` and `BorderRadius` components from the bundle anyways.

src/theme/widget.rs
```diff
@@ -47,8 +47,55 @@ pub fn label(text: impl Into<String>) -> impl Bundle {
...
+        ),
+    )
+}
+
+/// A small square button with text and an action defined as an [`Observer`].
+pub fn button_small<E, B, M, I>(text: impl Into<String>, action: I) -> impl Bundle
+where
+    E: Event,
+    B: Bundle,
+    I: IntoObserverSystem<E, B, M> + Sync,
```
benfrankel, line 78:
```suggestion
    I: IntoObserverSystem<E, B, M>,
```
The `+ Sync` isn't required here I believe? Same goes for `fn button`.

src/screens/settings.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn spawn_settings_screen(mut commands: Commands) {
+    commands.spawn((
+        widget::ui_root("Settings Screen"),
+        StateScoped(Screen::Settings),
+        children![
+            widget::header("Settings"),
+            (
+                Name::new("Settings"),
+                Node {
```
benfrankel, line 31: Maybe we should have a `widget::grid` somehow, since it seems we're now doing this in two (technically three) places. Although I'm not sure how it would support column alignment because that requires putting `justify_self` on each child node AFAIK. Ignore for this PR, but I'll give this some more thought later.

src/screens/settings.rs
```diff
@@ -0,0 +1,93 @@
...
+struct GlobalVolumeLabel;
+
+fn spawn_settings_screen(mut commands: Commands) {
+    commands.spawn((
+        widget::ui_root("Settings Screen"),
+        StateScoped(Screen::Settings),
+        children![
+            widget::header("Settings"),
+            (
+                Name::new("Settings"),
```
benfrankel, line 30:
```suggestion
                Name::new("Settings Grid"),
```

src/screens/settings.rs
```diff
@@ -0,0 +1,93 @@
...
+                                Node {
+                                    padding: UiRect::horizontal(Val::Px(10.0)),
+                                    justify_content: JustifyContent::Center,
+                                    ..default()
+                                },
+                                children![(widget::label(""), GlobalVolumeLabel)],
+                            ),
+                            widget::button_small("+", raise_volume),
+                        ],
+                    ),
```
benfrankel, line 63: Pull this section out to a `fn volume_widget() -> impl Bundle` IMO.

src/screens/settings.rs
```diff
@@ -0,0 +1,93 @@
...
+pub(super) fn plugin(app: &mut App) {
+    app.add_systems(OnEnter(Screen::Settings), spawn_settings_screen);
+    app.add_systems(
+        Update,
+        update_volume_label.run_if(in_state(Screen::Settings)),
+    );
+}
+
+#[derive(Component)]
+struct GlobalVolumeLabel;
```
benfrankel, line 21: Move this down to right above `fn update_volume_label`. Also derive `Reflect` and reflect `Component`, and add `app.register_type::<GlobalVolumeLabel>()`.

benfrankel (changes requested):

src/theme/widget.rs
janhohenheim: Yeah, but that way one could actually override something like the background color, which is neat :)
Will make it non-pub.

src/audio.rs
```diff
@@ -34,31 +20,3 @@ pub struct SoundEffect;
...
-
-fn update_music_for_new_global_scale(
-    music_volume_factor: Res<GlobalMusicVolumeScale>,
-    mut sinks: Query<(&mut AudioSink, &UnscaledVolume), With<Music>>,
-) {
-    for (mut sink, unscaled_volume) in &mut sinks {
-        let sink = sink.bypass_change_detection();
-        sink.set_volume(music_volume_factor.0 * unscaled_volume.0);
-    }
-}
```
SecretPocketCat: Smt. like this might still be needed based on the note from [GlobalVolume](https://docs.rs/bevy/latest/bevy/audio/struct.GlobalVolume.html).

> Note: Changing [GlobalVolume](https://docs.rs/bevy/latest/bevy/prelude/struct.GlobalVolume.html) does not affect already playing audio.

janhohenheim: heck

src/theme/widget.rs
benfrankel: YAGNI for now IMO. Could remove `BackgroundColor` from `button_base` as well.

janhohenheim: https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/5a9ecb876ee5934c80eb02cff86b5c26280b0427

src/screens/settings.rs
janhohenheim: https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/f049cd1e678d85dcd68fcfe9e936ca04e1649519

src/screens/settings.rs
janhohenheim: https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/93a7716ce668cdb3d40ed6ca286111f21d8fb8d6

src/audio.rs
janhohenheim: Opened an issue for this: https://github.com/bevyengine/bevy/issues/18952

src/theme/widget.rs
benfrankel: Oh, right, you can't just `(button(...), Node { ... })` because the actual button is a child entity. That's awkward. I thought we could remove `button_bundle` entirely as a consequence of this change. `SpawnObserver` would make this more ergonomic but alas.

janhohenheim: Should I go back to the `.insert` version? IMO it's less footgunny as it cannot ever result in duplicate components being accidentally spawned

src/screens/settings.rs
benfrankel: Oh I didn't mean the whole grid. Suppose a user of this template wants to add another settings widget to the grid for resolution or whatever. It'd be cleaner to encapsulate each widget as a bundle function and add that function as another child in the settings grid.

src/screens/settings.rs
```diff
@@ -0,0 +1,99 @@
...
+const MIN_VOLUME: f32 = 0.0;
+const MAX_VOLUME: f32 = 3.0;
+
+pub(super) fn plugin(app: &mut App) {
+    app.register_type::<GlobalVolumeLabel>();
+    app.add_systems(OnEnter(Screen::Settings), spawn_settings_screen);
+    app.add_systems(
+        Update,
+        update_volume_label.run_if(in_state(Screen::Settings)),
+    );
```
benfrankel, line 18:
```suggestion
    app.add_systems(OnEnter(Screen::Settings), spawn_settings_screen);
    
    app.register_type::<GlobalVolumeLabel>();
    app.add_systems(
        Update,
        update_volume_label.run_if(in_state(Screen::Settings)),
    );
```
Nit

benfrankel (changes requested):

src/theme/widget.rs
benfrankel: Yeah, the `.insert` version is better. FWIW the duplicate component footgun exists any time you do `(bundle_function(), MyComponent)` :sweat_smile:

janhohenheim: https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/b60bea33aa657c1072d4511a6e08a9b42c2f466d

src/screens/settings.rs
janhohenheim: Makes sense :)
https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/9cd8ebf8ebcd2706c2bd8c37de80c10f1a361353

src/screens/settings.rs
```diff
@@ -0,0 +1,100 @@
...
+//! We can add all manner of settings and accessibility options here.
+//! For 3D, we'd also place the camera sensitivity and FOV here.
+
+use bevy::{audio::Volume, prelude::*, ui::Val::*};
+
+use crate::{screens::Screen, theme::prelude::*};
+
+const MIN_VOLUME: f32 = 0.0;
+const MAX_VOLUME: f32 = 3.0;
+
```
benfrankel, line 11: I'd move these to just below `fn volume_widget()`. Not blocking.

src/theme/widget.rs
```diff
@@ -59,19 +106,11 @@ where
     (
         Name::new("Button"),
         Node::default(),
-        Children::spawn(SpawnWith(|parent: &mut ChildSpawner| {
+        Children::spawn(SpawnWith(move |parent: &mut ChildSpawner| {
```
benfrankel, line 109:
```suggestion
        Children::spawn(SpawnWith(|parent: &mut ChildSpawner| {
```
Is the `move` still needed?

src/audio.rs
```diff
@@ -12,9 +12,6 @@ pub fn music(handle: Handle<AudioSource>) -> impl Bundle {
     (AudioPlayer(handle), PlaybackSettings::LOOP, Music)
 }
 
-/// An organizational marker component that should be added to a spawned [`AudioPlayer`] if it's in the
-/// general "sound effect" category (e.g. footsteps, the sound of a magic spell, a door opening).
-///
```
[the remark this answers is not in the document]
janhohenheim: Whoops https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/99a9119af3ff484a56e419f95c67fb14f00094f8

src/screens/settings.rs
janhohenheim: https://github.com/TheBevyFlock/bevy_new_2d/pull/374/commits/f54a34db650721627b1d2e40219f5e01fba36d5c

src/lib.rs
```diff
@@ -43,12 +39,6 @@ impl Plugin for AppPlugin {
                     }
                     .into(),
                     ..default()
-                })
-                .set(AudioPlugin {
-                    global_volume: GlobalVolume {
-                        volume: Volume::Linear(0.3),
-                    },
-                    ..default()
```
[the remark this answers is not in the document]
janhohenheim: I removed it because it seems weird for the player to have the default audio be at 80%, and I didn't want to add the extra boilerplate of doing math with `const DEFAULT_GLOBAL_VOLUME: f32 = 0.3`

benfrankel (approved):
Looks good enough to merge now :tada:

We can continue to refine the settings menu in future PRs, maybe adding a pause menu that can navigate to settings as well.

PR #375, opened 2025-04-26
Replace leftover command-based spawning API

benfrankel:
As part of this PR I also edited a few comments, mostly by deleting things I considered to be over-explaining or confusing. I'm open to opinions on those changes.

### DIFF
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -4,8 +4,17 @@ use bevy::prelude::*;
 
 use crate::demo::player::{PlayerAssets, player};
 
+/// A system that spawns the main level.
+pub fn spawn_level(
+    mut commands: Commands,
+    player_assets: Res<PlayerAssets>,
+    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
+) {
+    commands.spawn(level(&player_assets, &mut texture_atlas_layouts));
+}
+
 /// The main level.
-pub fn level(
+fn level(
     player_assets: &PlayerAssets,
     texture_atlas_layouts: &mut Assets<TextureAtlasLayout>,
 ) -> impl Bundle {
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -2,7 +2,10 @@
 
 use bevy::prelude::*;
 
-use crate::demo::player::{PlayerAssets, player};
+use crate::{
+    demo::player::{PlayerAssets, player},
+    screens::Screen,
+};
 
 /// A system that spawns the main level.
 pub fn spawn_level(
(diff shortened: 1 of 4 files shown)


src/screens/gameplay.rs
```diff
@@ -22,8 +24,12 @@ pub(super) fn plugin(app: &mut App) {
...
 }
 
-fn spawn_level(mut commands: Commands) {
-    commands.queue(spawn_level_command);
+fn spawn_level(
+    mut commands: Commands,
+    player_assets: Res<PlayerAssets>,
+    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
+) {
+    commands.spawn(level(&player_assets, &mut texture_atlas_layouts));
```
janhohenheim, line 32: Can we remove this level of indirection and just call a `spawn_level` function inside `level.rs` directly? Otherwise, the gameplay screen needs to query for a lot of things that are implementation details of the level and not the screen, imo

janhohenheim (changes requested):

src/screens/gameplay.rs
benfrankel: IMO this is a direct consequence of the lack of dependency injection in the spawn API that we're using, and it's better to face it than try to hide it. I can see moving `fn spawn_level` into `level.rs` (in addition to `fn level`), though.

janhohenheim: yep, that would be nice

benfrankel: Done

src/demo/level.rs
```diff
@@ -2,19 +2,26 @@
...
-    // No setup required for this plugin.
-    // It's still good to have a function here so that we can add some setup
-    // later if needed.
+/// A system that spawns the main level.
+pub fn spawn_level(
+    mut commands: Commands,
+    player_assets: Res<PlayerAssets>,
+    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
+) {
+    commands.spawn(level(&player_assets, &mut texture_atlas_layouts));
```
janhohenheim, line 13: I understand the idea behind it, and I'm on board in principle, but in practice this functional split looks quite redundant in this case.
Could we eliminate the `level` function?

janhohenheim (changes requested):

src/demo/level.rs
benfrankel: I guess so. It just means you wouldn't be able to spawn the level outside of running this system, although to be fair, that's the only way we're doing it right now. If we want to go that way, I'd argue we should go one step further and remove `level.rs`, moving the system back to `screen/gameplay.rs`, since we're setting up the gameplay screen by simply spawning the player.

janhohenheim: I think keeping the logic for what goes into the level in `level.rs` is cleaner. The level of indirection from level to player is imo justified because most games will have more than one thing in the level.

benfrankel: One more thing, I don't think `fn player` should be adding `StateScoped(Screen::Gameplay)`. It should be on the level entity, but `level.rs` shouldn't be doing that either. So providing `pub fn level` and then having `gameplay.rs` spawn that bundle plus `StateScoped` makes the most sense IMO. For now I'll add `StateScoped` in `level.rs`.

janhohenheim: That's fair

janhohenheim (approved):

PR #381, opened 2025-04-28
Switch to `Swatinem/rust-cache`

BD103:
From https://github.com/TheBevyFlock/bevy_new_2d/issues/377#issuecomment-2835276190:

> I think it may be worth switching to [`Swatinem/rust-cache`](https://github.com/Swatinem/rust-cache) instead. It's [used by the Rust project](https://github.com/search?q=repo%3Arust-lang%2Frust+%22Swatinem%2Frust-cache%22+language%3AYAML&type=code) and provides more features than `Leafwing-Studios/cargo-cache`. Its strategy of only caching dependencies, disabling incremental compilation, and cleaning files over a week old is likely better than what we're currently doing.

This PR is a test to see how `Swatinem/rust-cache` performs in CI, and whether it is better than `Leafwing-Studios/cargo-cache` in practice or not. Note that we can [make the cache only save when run on `main`](https://github.com/Swatinem/rust-cache/blob/9d47c6ad4b02e050fd481d890b2ea34778fd09d6/README.md?plain=1#L69-L70), which should avoid creating needless caches [that cannot be used](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,9 +28,7 @@ jobs:
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
 
       - name: Populate target directory from cache
-        uses: Leafwing-Studios/cargo-cache@v2
-        with:
-          sweep-cache: true
+        uses: Swatinem/rust-cache@v2
 
       - name: Run tests
         run: |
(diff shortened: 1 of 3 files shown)


BD103:
> Note that we can [make the cache only save when run on `main`](https://github.com/Swatinem/rust-cache/blob/9d47c6ad4b02e050fd481d890b2ea34778fd09d6/README.md?plain=1#L69-L70), which should avoid creating needless caches [that cannot be used](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache).

Before merging this, I definitely want to enable `save-if`. (I can't yet, though, or else I wouldn't be able to test on this branch!)

janhohenheim (approved):

benfrankel (approved):
I suppose we should add `save-if` before merging this, and clear the current GitHub cache after merging? I'm happy to give `Swatinem/rust-cache` a trial period.

janhohenheim:
Cleared the cache :heavy_check_mark:

PR #383, opened 2025-04-29
Make web builds more robust

janhohenheim:
Otherwise, we might get this:

image

This error started cropping up in my 0.16 upgrades, dunno why. It happens on specific uses of `rand`, including through dependencies, so this might be very subtle for a user, especially since the error message is less then helpful.

Let's just remove that potential footgun :) I don't know how this interacts with the `RUSTFLAGS` env var that treats warnings as errors, but since we currently only run tests on native anyways, I figured it's alright.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ dev_native = [
 # Disable debug functionality in release builds.
 default_features = false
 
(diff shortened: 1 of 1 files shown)


benfrankel (approved):
I don't quite get it, but looks good to me.

PR #384, opened 2025-04-29
Enable track_location

janhohenheim:
This improved Bevy-emitted warnings.

E.g.

image

gives some more debugging information:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2024"
 license = "MIT OR Apache-2.0 OR CC0-1.0"
 
 [dependencies]
-bevy = { version = "0.16", features = ["wayland", "track_location"] }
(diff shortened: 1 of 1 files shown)


benfrankel (commented):
I just enabled this feature in my own template :)

Do we want this enabled in release builds, or should it be a `dev` feature?

janhohenheim:
Whoops!
I feel like it would be good practice to always activate this. If a player reports an issue, it would be nice to have this info in the console on web, and on native, I imagine it would be written into a log file.
But I'm fine with gating it to dev if you want :)

benfrankel:
Well I guess the question is the performance cost. I imagine there's some overhead involved. I'll opt for this being a `dev` feature for now, until / unless information arises that the overhead is negligible.

janhohenheim:
there ya go

benfrankel (approved):

PR #386, opened 2025-04-30
Install bevy_cli with cargo-binstall

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/330.

This will still compile from source until Bevy CLI is released to GitHub.


benfrankel:
I suppose this _is_ blocked on https://github.com/TheBevyFlock/bevy_cli/issues/354 then, although it honestly seems like a `cargo-binstall` bug.

benfrankel:
This should work now.

benfrankel:
Okay should be good now! This brings the release workflow from 27-30m down to 14-15m in `pyri_new_jam` :)

PR #389, opened 2025-05-01
Use `process` instead of `shell`

benfrankel:
We don't need any shell-specific features like glob, pipe, etc., we're just running a command with arguments, so we can use the simpler `"process"` task type.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -3,10 +3,10 @@
     "tasks": [
         {
             "label": "Run native dev",
-            "type": "shell",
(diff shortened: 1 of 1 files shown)


janhohenheim (approved):

janhohenheim:
What makes this more desirable over using shell?

benfrankel:
> What makes this more desirable over using shell?

Nothing in particular afaik. There's just no reason to spawn a shell that parses the command string and then starts the process with specified arguments, rather than starting the process with specified arguments directly.

PR #391, opened 2025-05-01
Don't include dev features in release

benfrankel:
`bevy build --release` is not the same as `bevy build --profile release` currently. If we want Bevy CLI to pick up our release configuration, we need to explicitly pass `--release`, at which point we might as well remove the `--profile` argument at the cost of some workflow flexibility that we aren't using anyways.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -106,27 +106,23 @@ jobs:
         include:
           - platform: web
             targets: wasm32-unknown-unknown
-            profile: web-release
             package_ext: .zip
             runner: ubuntu-latest
 
           - platform: linux
             targets: x86_64-unknown-linux-gnu
-            profile: release
             features: bevy/wayland
             package_ext: .zip
             runner: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -106,27 +106,23 @@ jobs:
         include:
           - platform: web
             targets: wasm32-unknown-unknown
-            profile: web-release
```
janhohenheim, line 109: To ensure this still does the same thing, we need to move the profile settings in `Cargo.toml` to `[package.metadata.bevy_cli.web.release]`

janhohenheim (changes requested):

.github/workflows/release.yaml
benfrankel: I'm not sure what you mean. `bevy build --release web` will pass `--profile web-release` to `cargo`.

janhohenheim: oh, I see.

janhohenheim (approved):

PR #392, opened 2025-05-04
Fix interaction bugs

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/380.

This fixes both the SFX _and_ the click/drag bugs described in that issue.


src/theme/interaction.rs
```diff
@@ -48,30 +45,45 @@ struct InteractionAssets {
...
     }
 }
 
-fn trigger_interaction_sound_effect(
-    interaction_query: Query<&Interaction, Changed<Interaction>>,
-    interaction_assets: Res<InteractionAssets>,
+fn play_on_hover_sound_effect(
+    trigger: Trigger<Pointer<Over>>,
+    mut commands: Commands,
+    interaction_assets: Option<Res<InteractionAssets>>,
```
janhohenheim, line 64: Why is this an `Option`?

src/theme/widget.rs
```diff
@@ -24,6 +24,7 @@ pub fn ui_root(name: impl Into<Cow<'static, str>>) -> impl Bundle {
             row_gap: Px(20.0),
             ..default()
         },
+        Pickable::IGNORE,
```
janhohenheim, line 27: Can we add a comment about why this is needed?

src/theme/interaction.rs
benfrankel: The system this was converted from had `.run_if(resource_exists::<InteractionAssets>)`.

janhohenheim: Isn't that more elegant?

src/theme/widget.rs
benfrankel: Added a comment.

benfrankel: Oh wait nvm, this is on a different line. Let me add another comment.

janhohenheim (approved):

PR #395, opened 2025-05-06
Tweak workflow timeouts

janhohenheim:
The first cacheless run can often take longer than 30 mins according to reports on discord.
I also moved the env vars together :)

Note that I'm experimenting with using a single shared key for the cache in Foxtrot, I'll report back on how well that works.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,7 +98,7 @@ jobs:
   format:
     name: Format
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 10
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


benfrankel:
Rather than remove all the timeouts, I'd increase the timeout on build for web and tests to 60 minutes. The other jobs should be fast because they don't have to compile your game.

janhohenheim:
Tweaked them all a bit

.github/workflows/ci.yaml
```diff
@@ -97,7 +98,7 @@ jobs:
   format:
     name: Format
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 5
```
benfrankel, line 101:
```suggestion
    timeout-minutes: 10
```
This _should_ be fast, but with private runners and CPU RNG I'd go a little higher here.

benfrankel (approved):

PR #396, opened 2025-05-08
Reduce workflow cache size

benfrankel:
Passing `debuginfo=0` would reduce the cache size even further (order of ~200MB), but may remove some debug information we actually want in CI. To be on the safe side, this PR uses `debuginfo=1` which should still be a huge win.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: --deny warnings
+  RUSTFLAGS: --deny warnings -C debuginfo=1
   RUSTDOCFLAGS: --deny warnings
   LINTER_TOOLCHAIN: nightly-2025-04-03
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -86,13 +86,13 @@ jobs:
...
 
-      - name: Populate target directory from cache
+      - name: Restore Rust cache
         uses: Swatinem/rust-cache@v2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Run clippy lints
-        run: cargo clippy --locked --workspace --all-targets --all-features -- --deny warnings
+        run: cargo clippy --locked --workspace --all-targets --all-features --no-deps
```
janhohenheim, line 95: What does this do for clippy? I don't get warnings from dependency code anyways.

benfrankel: I'm not sure how real it is, but `cargo clippy -h` says this:
```
    --no-deps                Run Clippy only on the given crate, without linting the dependencies
```

benfrankel: It may be for excluding dependencies that are members of the workspace? In which case I guess we _don't_ want to pass this argument.

benfrankel: Context: https://github.com/rust-lang/rust-clippy/pull/6188. I'll remove this argument.

janhohenheim (approved):

BD103:
[Looking at the docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#debuginfo), I think we can even get away with `debuginfo=line-tables-only` if we only care about backtraces. `debuginfo=1` seems to include extra information that's unused by backtraces (https://github.com/rust-lang/rust/issues/64405), which is why `line-tables-only` was added.

PR #398, opened 2025-05-08
Move level music to level

janhohenheim:
Reasoning:
- The level can decide its own music
- The `LevelAssets` resource is useful in practice, as it will contain the level itself (in e.g. a `Handle<Scene>`) in a real project
- We remove an abstraction (the resource that holds an entity)
- fewer functions overall
- The `spawn_level` function looks less sad :)

### DIFF
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -42,12 +42,12 @@ pub fn spawn_level(
         Transform::default(),
         Visibility::default(),
         StateScoped(Screen::Gameplay),
-        children![player(400.0, &player_assets, &mut texture_atlas_layouts)],
-    ));
-
-    commands.spawn((
-        Name::new("Level Music"),
-        StateScoped(Screen::Gameplay),
-        music(level_assets.music.clone()),
+        children![
+            player(400.0, &player_assets, &mut texture_atlas_layouts),
+            (
+                Name::new("Gameplay Music"),
+                music(level_assets.music.clone())
+            )
+        ],
     ));
 }
--- a/src/demo/level.rs
+++ b/src/demo/level.rs
@@ -14,28 +14,9 @@ pub(super) fn plugin(app: &mut App) {
     app.load_resource::<LevelAssets>();
 }
 
-/// A system that spawns the main level.
-pub fn spawn_level(
-    mut commands: Commands,
-    player_assets: Res<PlayerAssets>,
-    mut texture_atlas_layouts: ResMut<Assets<TextureAtlasLayout>>,
-) {
-    commands.spawn((
-        Name::new("Level"),
-        Transform::default(),
-        Visibility::default(),
-        StateScoped(Screen::Gameplay),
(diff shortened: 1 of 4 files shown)


src/demo/level.rs
```diff
@@ -3,14 +3,38 @@
...
 };
 
+pub(super) fn plugin(app: &mut App) {
+    app.register_type::<LevelAssets>();
+    app.load_resource::<LevelAssets>();
+}
+
+#[derive(Resource, Asset, Clone, Reflect)]
+#[reflect(Resource)]
+pub struct LevelAssets {
```
janhohenheim, line 19: needs to be `pub` because `spawn_level` uses it, which itself is used across modules.

src/demo/level.rs
```diff
@@ -20,4 +44,10 @@ pub fn spawn_level(
         StateScoped(Screen::Gameplay),
         children![player(400.0, &player_assets, &mut texture_atlas_layouts)],
     ));
+
+    commands.spawn((
+        Name::new("Level Music"),
+        StateScoped(Screen::Gameplay),
+        music(level_assets.music.clone()),
```
janhohenheim, line 51: Could also be a component on the level itself tbh, but I think it's slightly neater this way

benfrankel: I'd prefer this as a child entity of the level, and then you can omit `StateScoped`. Also, now that I see this, I'm wondering why we're doing e.g. `stop_credits_music` when we could just put `StateScoped` on the music entity (and `Name`).

src/demo/mod.rs
```diff
@@ -11,5 +11,10 @@ mod movement;
 pub mod player;
 
 pub(super) fn plugin(app: &mut App) {
-    app.add_plugins((animation::plugin, movement::plugin, player::plugin));
+    app.add_plugins((
+        animation::plugin,
+        movement::plugin,
+        player::plugin,
+        level::plugin,
```
benfrankel, line 18:
```suggestion
        level::plugin,
        movement::plugin,
        player::plugin,
```

src/demo/level.rs
```diff
@@ -3,14 +3,38 @@
...
+            music: assets.load("audio/music/Fluffing A Duck.ogg"),
+        }
+    }
+}
+
 /// A system that spawns the main level.
 pub fn spawn_level(
     mut commands: Commands,
     player_assets: Res<PlayerAssets>,
+    level_assets: Res<LevelAssets>,
```
benfrankel, line 37:
```suggestion
    level_assets: Res<LevelAssets>,
    player_assets: Res<PlayerAssets>,
```
Subjective :P

benfrankel (changes requested):

benfrankel:
One thing to consider is, if you have multiple levels, does this mean we're prescribing that the music will stop and new music will start when you progress to the next level? I think either way is valid and users will change this code to their liking anyways so it doesn't really matter.

benfrankel (approved):

PR #405, opened 2025-05-12
Boost CI performance

janhohenheim:


benfrankel:
Related: https://github.com/TheBevyFlock/bevy_new_2d/issues/119.

.cargo/config_ci.toml
```diff
@@ -0,0 +1,53 @@
+# This config is used for the CI workflow.
```
janhohenheim, line 1: Could also use this for release, I believe

.github/workflows/ci.yaml
```diff
@@ -23,7 +23,21 @@ jobs:
...
+        uses: rui314/setup-mold@v1
+
+      - name: Active CI cargo config
+        run: mv .cargo/config_ci.toml .cargo/config.toml
+
+      - name: Install cargo binstall
+        uses: cargo-bins/cargo-binstall@main
+
+      - name: Install nextest
+        run: cargo binstall cargo-nextest --secure --force
```
janhohenheim, line 40: Note that without `--force`, it somehow didn't land in the PATH?

.github/workflows/ci.yaml
```diff
@@ -34,40 +48,41 @@ jobs:
...
 
       - name: Restore Rust cache
         uses: Swatinem/rust-cache@v2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
-      - name: Install Bevy CLI
-        run: cargo install --git='https://github.com/TheBevyFlock/bevy_cli' --locked bevy_cli
+      - name: Check web build
+        run: RUSTFLAGS='--deny warnings -C debuginfo=line-tables-only --cfg getrandom_backend="wasm_js"' cargo check --config 'profile.web.inherits="dev"' --features dev --no-default-features --profile web --target wasm32-unknown-unknown
```
janhohenheim, line 84: the `--cfg` part didn't end up being used when putting this into the job's `env`, for some reason

.cargo/config_ci.toml
```diff
@@ -0,0 +1,53 @@
...
+# Feel free to also use it locally for fast builds if you have the required setup!
+
+[unstable]
+codegen-backend = true
+
+[profile.dev]
+codegen-backend = "cranelift"
+
+# compile deps with llvm still for runtime perf
```
janhohenheim, line 10: This comment is from my `~/.cargo/config.toml`, so that doesn't make sense, sry.
However, I have experienced the test runner crash before when compiling everything with cranelift. Not sure if that's still the case, but I'd be careful about removing this line.

.github/workflows/ci.yaml
janhohenheim: Could also move it to the `config_ci.toml`

benfrankel (approved):
Approving for followup in another PR.

PR #407, opened 2025-05-13
Fix ignored rustflags

janhohenheim:


benfrankel (approved):
We can merge this, but I'm still curious about the `LINTER_TOOLCHAIN` environment variable.

PR #408, opened 2025-05-13
Optimize CI speed by a tiny bit

janhohenheim:
- Cache dynamic libraries
- Compile tests with fewer optimizations
- Disable [concurrent CI runs](https://stackoverflow.com/a/72408109/5903309) on the same branch, PR, or tag
  - This should save computation on private repos

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-    
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
benfrankel, line 10:
```suggestion
  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
```
See comments from `BjornO` in the stackoverflow thread you linked. Including `github.run_id` as a fallback may be overkill here since we don't have a `workflow_dispatch` trigger, but I figure why not.

.github/workflows/ci.yaml
```diff
@@ -36,18 +40,25 @@ jobs:
...
+        run: |
+          # Setting LD_LIBRARY_PATH is a workaround for <https://github.com/TheBevyFlock/bevy_new_2d/pull/318#issuecomment-2585935350>.
+          echo "LD_LIBRARY_PATH=$(rustc --print target-libdir)" >> $GITHUB_ENV
+          echo "ld_library_path=$(rustc --print target-libdir)" >> $GITHUB_OUTPUT
+
       - name: Restore Rust cache
         uses: Swatinem/rust-cache@v2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
+          cache-directories: ${{ steps.ld_library_path.outputs.ld_library_path }}
```
benfrankel, line 54: Is `env` not allowed here, for `${{ env.LD_LIBRARY_PATH }}`?

janhohenheim: It is! Good idea :)

.github/workflows/ci.yaml
```diff
@@ -27,27 +31,30 @@ jobs:
...
+        uses: taiki-e/install-action@nextest
 
       - name: Install dependencies
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
 
+      - name: Set LD_LIBRARY_PATH
+        id: ld_library_path
+        run: |
+          # Setting LD_LIBRARY_PATH is a workaround for <https://github.com/TheBevyFlock/bevy_new_2d/pull/318#issuecomment-2585935350>.
+          echo "LD_LIBRARY_PATH=$(rustc --print target-libdir)" >> $GITHUB_ENV
```
benfrankel, line 44:
```suggestion
        run: |
          # Setting LD_LIBRARY_PATH is a workaround for <https://github.com/TheBevyFlock/bevy_new_2d/pull/318#issuecomment-2585935350>.
          echo "LD_LIBRARY_PATH=$(rustc --print target-libdir)" >>"${GITHUB_ENV}"
```
I'm still curious what effect this has on CI time if any, but happy to merge it for now after this nit.

benfrankel (changes requested):

benfrankel (approved):

PR #414, opened 2025-05-16
Update release profile options

benfrankel:
Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/402.

We're mentioning `oxipng` and `gifski` in `docs/tooling.md` now, so all that remains from that discussion is to experiment with PGO, BOLT, and `upx`, but there's a separate issue for PGO and BOLT (https://github.com/TheBevyFlock/bevy_new_2d/issues/413) and `upx` could be moved to a separate issue as well.

We may want to update the bevy website's release optimization section based on the data analyzed in that thread as well, but that's not a `bevy_new_2d` issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ dev = [
     "bevy/dynamic_linking",
     "bevy/bevy_dev_tools",
     "bevy/bevy_ui_debug",
-    # Improve error messages coming from Bevy
+    # Improve error messages coming from Bevy.
     "bevy/track_location",
 ]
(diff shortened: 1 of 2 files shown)


janhohenheim:
I think we should have the same release optimizations as the guide to not send mixed signals

benfrankel:
> I think we should have the same release optimizations as the guide to not send mixed signals

Do you mean the suggestions should be updated upstream before merging this PR?

janhohenheim:
yep

PR #418, opened 2025-05-22
Streamline hot-patching with `subsecond`

benfrankel:
Also updated our other details section to use this styling when appropriate:

2025-05-22_1747952917_222x39


README.md
```diff
@@ -55,28 +55,51 @@ This template also comes with [VS Code tasks](./.vscode/tasks.json) and [JetBrai
...
+<details>
+  <summary><ins>(Optional) Improving compile times</ins></summary>
 
+  [`.cargo/config_fast_builds.toml`](./.cargo/config_fast_builds.toml) contains documentation on how to set up your environment to improve compile times.
+  After you've fiddled with it, rename it to `.cargo/config.toml` to enable it.
 </details>
 
 <details>
-    <summary>(Optional) Improve your compile times</summary>
+  <summary><ins>(Optional) Hot-patching with Dioxus</ins></summary>
```
janhohenheim, line 80:
```suggestion
  <summary><ins>(Optional) Hot-patching with `subsecond`</ins></summary>
```

README.md
```diff
@@ -55,28 +55,51 @@ This template also comes with [VS Code tasks](./.vscode/tasks.json) and [JetBrai
...
+
+  Follow the instructions in [`bevy_simple_subsecond_system`](https://github.com/TheBevyFlock/bevy_simple_subsecond_system/) to
```
janhohenheim, line 82: I'd add a little sentence about what hot-patching is. Something like "With hot-patching, you can edit the game's code while it's running and see the effects live" or so

janhohenheim (changes requested):
Looks nearly perfect :)

README.md
benfrankel:
```suggestion
  <summary><ins>(Optional) Hot-patching with <code>subsecond</code></ins></summary>
```

README.md
benfrankel: Done.

README.md
```diff
@@ -55,28 +55,53 @@ This template also comes with [VS Code tasks](./.vscode/tasks.json) and [JetBrai
...
+
+  To set this up, follow the instructions in [`bevy_simple_subsecond_system`](https://github.com/TheBevyFlock/bevy_simple_subsecond_system/).
+  Make sure to read the [`Known Limitations`](https://github.com/TheBevyFlock/bevy_simple_subsecond_system/?tab=readme-ov-file#known-limitations)
+  section and comment out the `bevy/dynamic_linking` feature in [`Cargo.toml`](./Cargo.toml).
+
+  Annotate your systems to enable hot-patching.
+  The functions they call can be hot-patched too; no additional annotations required!
+
+  ```rust
+  #[cfg_attr(feature = "dev_native", hot)]
```
janhohenheim, line 93: We should make the assumption explicit here: make `bevy_simple_subsecond_system` an optional dependency enabled on `dev_native`.
Maybe a little `Cargo.toml` snippet would be nice. I think GitHub can correctly render `diff`s:

```diff
# Cargo.toml

[dependencies]
+ bevy_simple_subsecond_system = { version = "0.1", optional = true }

[features]
dev_native = [
  "bevy/file_watcher",
+ "bevy_simple_subsecond_system",
]
```

benfrankel: I think _technically_ it would only increase compile times if a user doesn't do this, right? I can add the snippet though.

janhohenheim: No, it would break wasm builds

janhohenheim: Let me fix that upstream, actually

janhohenheim, line 93: Could in theory remove this `cfg` now, at the cost of some extra compile time and a bigger binary

janhohenheim: Imo either we remove this feature gate OR we mention how to make the dependency entirely optional. Otherwise, people might pull it in without a feature gate and then still gate the `#[hot]` attr, which would be useless

janhohenheim (approved):

PR #420, opened 2025-05-24
Use tips from perf book

janhohenheim:


benfrankel:
I want to look into this a bit more before merging, even though it is what they seem to be recommending. Also notably, public GitHub runners have 4 CPUs and private ones have 2 CPUs.

janhohenheim:
Since they have only <4 cores, this doesn't make sense

PR #425, opened 2025-05-27
Support pausing gameplay

benfrankel:
Changes:
- Add `Pause(bool)` state in `src/pause.rs` to represent whether the game is paused.
- Add `PausableSystems` system set in `src/pause.rs` to easily mark systems that should _not_ run when the game is paused. It's important that this is granular and opt-in.
- Make P key press fully close the current menu while in the gameplay screen (so if you press P while in the settings menu, it will close the menu completely, instead of doing nothing. pressing escape still only goes back 1 level)
- Spawn a pause overlay that is scoped to `Pause(true)`.

Questions for reviewers:
- Is `src/pause.rs` a reasonable place / name for this? I think it's the most straightforward option, but maybe a little confusing with `src/menu/pause.rs`. Technically that's a whole different thing: it's strictly a menu that happens to be called the "pause menu".
- How do we feel about the fact that `GlobalZIndex(2)` is now required for every menu's root node?

But I'm not _too_ concerned about the code quality; I think it's reasonable and we can improve it over time.

Here's how it looks:

2025-05-27_1748400836_1920x1080


janhohenheim (changes requested):
- Implementation looks great!
- Not sure why we need two different Z indices at all? Can't the background color be part of the pause UI?
- I'd rename `src/pause.rs` to `pause_systems` or similar

janhohenheim:
We could also move the Z indices into consts, but not blocking on that :)

benfrankel:
> Also, this doesn't pause music yet, right? That's also fine.

Yes, I'm not 100% if we want to or not.

PR #431, opened 2025-05-29
Remove lockfile command footgun

janhohenheim:


benfrankel:
I think we should keep this using `include` instead of `exclude`. "Skipped" means `cargo-generate` copies the file directly with no changes, and `Done` means `cargo-generate` interprets the file as a Liquid template and looks for `{...}` tags to translate. We don't want any of our files to be interpreted as Liquid templates _except_ for our `.template` files.

benfrankel (approved):

PR #438, opened 2025-06-11
Switch to manual release only

benfrankel:
Make releases more intentional, and put the main configuration options front and center.

I do something similar in `pyri_new_jam`, but will have to test this in `bevy_new_2d` specifically after this is merged.

Alternative to https://github.com/TheBevyFlock/bevy_new_2d/pull/436.
Relates to https://github.com/TheBevyFlock/bevy_new_2d/issues/212 and https://github.com/TheBevyFlock/bevy_new_2d/issues/292.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,26 +10,26 @@ on:
         type: string
       build_for_windows:
         description: "Build for Windows? (WARNING: very slow!)"
-        default: false
+        default: true
         type: boolean
       build_for_macos:
         description: "Build for macOS?"
-        default: false
+        default: true
         type: boolean
       build_for_linux:
         description: "Build for Linux?"
-        default: false
+        default: true
         type: boolean
       build_for_web:
         description: "Build for web?"
-        default: false
+        default: true
         type: boolean
       upload_to_github:
         description: "Upload to GitHub releases?"
         default: true
         type: boolean
       upload_to_itch:
-        description: "Upload to itch.io?"
+        description: "Upload to itch.io? (if configured)"
         default: true
         type: boolean
       deny_warnings:
--- a/.github/workflows/release.yaml.template
+++ b/.github/workflows/release.yaml.template
@@ -10,26 +10,26 @@ on:
         type: string
       build_for_windows:
         description: "Build for Windows? (WARNING: very slow!)"
-        default: false
+        default: true
         type: boolean
       build_for_macos:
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -1,19 +1,43 @@
...
+        default: true
+        type: boolean
+      upload_to_itch:
+        description: "Upload to itch.io?"
+        default: true
+        type: boolean
+      deny_warnings:
+        description: "Deny warnings?"
+        default: false
+        type: boolean
```
BD103, line 38: I personally find it a bit strange that this is an option, as it doesn't seem to "fit" with the other options. I'm not sure what to do about it, though, and it's probably fine staying

.github/workflows/release.yaml
```diff
@@ -1,19 +1,43 @@
...
+      build_for_linux:
+        description: "Build for Linux?"
+        default: false
+        type: boolean
+      build_for_web:
+        description: "Build for web?"
+        default: false
+        type: boolean
```
BD103, line 26: I don't think these should all be disabled by default. If a user doesn't change any of the default settings, nothing will happen! I think some, if not all, of these should be enabled.

BD103 (changes requested):

.github/workflows/release.yaml
benfrankel: That's true. It's hard to say which ones should be enabled by default, though.
- Web only makes sense as a "neutral" target, but a lot of games target native only.
- Windows only makes sense as the most common target, but it's also the slowest to build.
- Enabling everything by default would better match the status quo before this PR, but that means defaulting to maximum CI minutes.

.github/workflows/release.yaml
```diff
@@ -1,19 +1,43 @@
...
+      upload_to_github:
+        description: "Upload to GitHub releases?"
+        default: true
+        type: boolean
+      upload_to_itch:
+        description: "Upload to itch.io?"
+        default: true
```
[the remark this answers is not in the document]
benfrankel: I changed the description to `"Upload to itch.io? (if configured)"`.

janhohenheim (approved):
Thought this through in my head once more and yes, this is the way to go IMO.
Thanks for implementing :)

benfrankel:
I'm going to merge this because it's an improvement over the current version and addresses several post-jam complaints. We can work out more of the details in future PRs.

PR #449, opened 2025-08-01
Update Cargo.toml

zeddash:
Making it clear the second part is to be used in conjunction with the first. I didn't know the syntax and thought it was A or B, not A or (A and B).

I got really confused and just an extra word "additional" would have cleared things up and I'd have known I need both.

I read the "Alternatively" as "that one has a rustflag check, this one doesn't"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ tracing = { version = "0.1", features = [
 # In addition to enabling the `wasm_js` feature, you need to include `--cfg 'getrandom_backend="wasm_js"'`
 # in your rustflags for both local and CI/CD web builds, taking into account that rustflags specified in
 # multiple places are NOT combined (see <https://github.com/rust-lang/cargo/issues/5376>).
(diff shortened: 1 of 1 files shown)


zeddash:
Wouldn't that then mean you just need the second and not both? It only started to work when I enabled both
```
[target.wasm32-unknown-unknown.dependencies]
getrandom = { version = "0.3", features = ["wasm_js"] }
```
and
```
[patch.crates-io]
getrandom = { git = "https://github.com/benfrankel/getrandom" }
```
But I was reading it that I could just use the patch

benfrankel:
Yes, in either case you have to add `getrandom` as an explicit dependency in order to specify the `wasm_js` feature on it. "Alternatively" only relates to the `rustflags` section. Maybe it would help to put a newline between the two sections.

zeddash:
I think just adding my tiny change of specifying the second part is additional is enough, just clearly states it’s, well, additional

________________________________

PR #451, opened 2025-08-11
Run `bevy_lint` v0.4.0 in CI

BD103:
In #445 a pinned `bevy_lint` version was used because the previous v0.3.0 version didn't support `let`-chains. Now that v0.4.0 is released, this is no longer necessary! (Plus, there have been several improvements to the action between the pinned commit and v0.4.0's release.)

I enabled the optional caching on `main` for the `bevy_lint` and `bevy_lint_driver` executables, which should speed up CI times at the miniscule cost of ~1.5 MB in cache for all 3 operating systems.

image

Definitely also check out the [release page for the list of changes](https://github.com/TheBevyFlock/bevy_cli/releases/tag/lint-v0.4.0)! There's been some cool progress since v0.3.0!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,7 +123,10 @@ jobs:
         run: sudo apt-get update; sudo apt-get install --no-install-recommends libasound2-dev libudev-dev libwayland-dev
 
       - name: Install Bevy Lint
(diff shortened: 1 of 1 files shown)


BD103:
> Nice! Is the toolchain env var still accurate?

It is! You updated it when you pinned the commit, and we haven't changed it between then and v0.4.0.

PR #454, opened 2025-09-23
Add fixes for wasm builds with wasm_js on release pipeline

jbuehler23:
Fix for getting the Release pipeline working that I had to make on my local workflow to get itch integration working. These fixes have already been temporarily added into the CI pipeline, they just needed to be extended to the release workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -220,10 +220,12 @@ jobs:
 
       - name: Install Bevy CLI
         if: ${{ env.is_platform_enabled == 'true' }}
-        run: cargo binstall --locked --no-confirm --force --git='https://github.com/TheBevyFlock/bevy_cli' bevy_cli
+        run: cargo install bevy_cli --git https://github.com/TheBevyFlock/bevy_cli
 
       - name: Build and add web bundle to app (Web)
         if: ${{ env.is_platform_enabled == 'true' && matrix.platform == 'web' }}
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -220,10 +220,12 @@ jobs:
 
       - name: Install Bevy CLI
         if: ${{ env.is_platform_enabled == 'true' }}
-        run: cargo binstall --locked --no-confirm --force --git='https://github.com/TheBevyFlock/bevy_cli' bevy_cli
+        run: cargo install bevy_cli --git https://github.com/TheBevyFlock/bevy_cli
 
       - name: Build and add web bundle to app (Web)
         if: ${{ env.is_platform_enabled == 'true' && matrix.platform == 'web' }}
+        env:
+          CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUSTFLAGS: '--cfg getrandom_backend="wasm_js"'
```
benfrankel, line 228: I would prefer the approach taken here to extend `RUSTFLAGS`:
https://github.com/TheBevyFlock/bevy_new_2d/blob/606336b019aa3fc6b01a12f57845bd2276612ffa/.github/workflows/ci.yaml#L165-L166
That way we won't have to install Bevy CLI `main` from source.

benfrankel (changes requested):
This change will have to be repeated in `.github/workflows/release.yaml.template` as well. That's the version that ends up in users' projects, whereas `release.yaml` is used to dogfood releasing the template itself.

benfrankel:
Yep, the release workflow is working now: https://github.com/TheBevyFlock/bevy_new_2d/actions/runs/17963944467.

PR #463, opened 2025-10-06
Fixed inputs context check against strings when they are booleans

Fjalar:
The upload to github releases job does not work currently due to the variable in the ´inputs' context being a boolean and not equal to ´true´. This changes the checks for both ´inputs.upload_to_github` and `inputs.deny_warnings`. Should fix https://github.com/TheBevyFlock/bevy_new_2d/issues/461#issue-3484755689


benfrankel (approved):

Fjalar:
Thanks for fixing the stuff i misunderstood and approving my first ever PR :)

PR #465, opened 2025-10-07
Deploy to Github Pages

Fjalar:
Implements https://github.com/TheBevyFlock/bevy_new_2d/issues/143

Adds the option to deploy to github pages. Might require enabling publishing pages through github actions in the repo settings as well as adding the deploying branch to the appropriate environment. Perhaps relevant for the documentation?


.github/workflows/release.yaml
```diff
@@ -32,6 +32,10 @@ on:
         description: "Upload to itch.io (if configured)"
         default: true
         type: boolean
+      deploy-to-pages:
+        description: "Deploy to Github Pages (requires web build)"
```
benfrankel, line 36:
```suggestion
      upload_to_github_pages:
        description: "Upload web build to GitHub Pages"
```

The `deploy` verb is good too, but I'd like to keep it consistent. We can either rename this, or rename `upload_to_itch` -> `deploy_to_itch`.
Should use underscores instead of hyphens either way. Also will need to make this change in other places to match.

.github/workflows/release.yaml
```diff
@@ -306,6 +310,13 @@ jobs:
           name: package-${{ matrix.platform }}
           retention-days: 1
 
+      - name: Upload web build artifact for Github Pages deployment
```
benfrankel, line 313:
```suggestion
      - name: Upload binary to workflow artifacts for GitHub Pages (web)
```
It's interesting we only upload the binary here, and not the entire package with `index.html` etc., is that intentional?

.github/workflows/release.yaml
```diff
@@ -351,3 +362,23 @@ jobs:
               tmp/"${channel}"/* \
               '${{ env.itch_page }}':"${channel#package-}"
           done
+
+  # Deploy to Github Pages
```
benfrankel, line 366:
```suggestion
  # Upload to GitHub Pages.
```

.github/workflows/release.yaml
```diff
@@ -351,3 +362,23 @@ jobs:
...
+    permissions:
+      pages: write # to deploy to Pages
+      id-token: write # to verify the deployment originates from an appropriate source
+
+    # Deploy to the github-pages environment
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-latest
```
benfrankel, line 380:
```suggestion
    runs-on: ubuntu-latest
    needs: build
    permissions:
      # Required to upload to GitHub Pages.
      pages: write
      # Required to verify the upload originates from an appropriate source.
      id-token: write
    environment:
      name: github-pages
      url: ${{ steps.pages.outputs.page_url }}

```
I don't understand the comment on the `id-token` permission.

I'm wondering what `name: github-pages` in `environment` is needed for. I see in the example in the action's [readme](https://github.com/actions/upload-pages-artifact) that they do this too, but the action also has an _input_ called `name` that already defaults to `github-pages` and is not required.

.github/workflows/release.yaml
```diff
@@ -306,6 +310,13 @@ jobs:
           name: package-${{ matrix.platform }}
           retention-days: 1
 
+      - name: Upload web build artifact for Github Pages deployment
+        if: ${{ env.is_platform_enabled == 'true' && matrix.platform == 'web' && inputs.deploy-to-pages }}
+        id: deployment
```
benfrankel, line 315:
```suggestion
        id: pages
```
`deployment` is not descriptive enough in the context of the full `release.yaml`.

.github/workflows/release.yaml
```diff
@@ -351,3 +362,23 @@ jobs:
...
+
+    # Deploy to the github-pages environment
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
```
benfrankel, line 383:
```suggestion
```
I know the example has this too, but I don't think we use this anywhere?

benfrankel (changes requested):
I did a style pass, but the content looks good, assuming that it works (which we can test after merging).

.github/workflows/release.yaml
benfrankel: Okay I was looking at the wrong action. The [`deploy-pages`](https://github.com/actions/deploy-pages) action doesn't have an input called `name`, and it does need this environment variable (although I'm not sure what it does).

.github/workflows/release.yaml
benfrankel: Hm, `page_url` is actually an output of the [`deploy-pages`](https://github.com/actions/deploy-pages) action, not the [`upload-pages-artifact`](https://github.com/actions/upload-pages-artifact) action. So it seems like we actually do need the `id` here but we don't need the `id` for the upload step?

Also, I'm confused how `environment` can reference the output of a step that hasn't run yet :question:

.github/workflows/release.yaml
benfrankel: I found an [issue](https://github.com/actions/deploy-pages/issues/329) about the `id-token` permission, it looks like a lot of people are confused about it. We should probably link to the issue in the comment.

.github/workflows/release.yaml
Fjalar: It does upload the entire package with html etc in my testing, it's just we don't want the containing folder to uploaded. The regular web build is a bit nested, resulting in the page becoming available at for example TheBevyFlock.github.io/bevy_new_2d/bevy_new_2d, with the sub-directory being the repo name and the sub-sub-directory being the binary name (I think).

Fjalar:
I'll look into updating the docs as well

.github/workflows/release.yaml
```diff
@@ -306,6 +310,13 @@ jobs:
           name: package-${{ matrix.platform }}
           retention-days: 1
 
+      - name: Upload binary to workflow artifacts for GitHub Pages (web)
```
benfrankel, line 313:
```suggestion
      - name: Upload web bundle to workflow artifacts for GitHub Pages (web)
```
Okay I see now, `bevy build web --bundle` gives the bundle the same name as the `cargo build` binary name, so here you're uploading the whole web bundle as an artifact 👍

Some thoughts out of scope for this PR: I wonder if we should be renaming the bundle folder to something else (like `app_binary_name` at least), but I suppose unlike the packages for other platforms, players don't download and unzip the package themselves so they never see the names.

.github/workflows/release.yaml
```diff
@@ -351,3 +362,20 @@ jobs:
...
           done
+
+  # Upload to GitHub Pages.
+  deploy-to-pages:
+    runs-on: ubuntu-latest
+    needs: build
+    permissions:
+      # Required to upload to GitHub Pages.
+      pages: write
+      # Required to verify the upload originates from an appropriate source.
```
benfrankel, line 373:
```suggestion
      # Required for GitHub Pages to verify the source of the upload.
      # See: <https://github.com/actions/deploy-pages/issues/329>.
```

.github/workflows/release.yaml
```diff
@@ -351,3 +362,20 @@ jobs:
...
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.pages.outputs.page_url }}
+
+    steps:
+      - name: Deploy to GitHub Pages
+        uses: actions/deploy-pages@v4
```
benfrankel, line 381:
```suggestion
      url: ${{ steps.deploy.outputs.page_url }}

    steps:
      - name: Deploy to GitHub Pages
        id: deploy
        uses: actions/deploy-pages@v4
```
The `${{ steps.*.outputs.page_url }}` actually comes from _this_ step, so this step needs the ID, sorry about that 😅

.github/workflows/release.yaml
```diff
@@ -351,3 +362,20 @@ jobs:
               tmp/"${channel}"/* \
               '${{ env.itch_page }}':"${channel#package-}"
           done
+
+  # Upload to GitHub Pages.
+  deploy-to-pages:
+    runs-on: ubuntu-latest
+    needs: build
+    permissions:
+      # Required to upload to GitHub Pages.
```
benfrankel, line 371:
```suggestion
  # Deploy to GitHub Pages.
  deploy-to-github-pages:
    runs-on: ubuntu-latest
    needs: build
    permissions:
      # Required to deploy to GitHub Pages.
```
IIUC now, I believe this job only deploys what we already uploaded earlier during the web build, so "deploy" was the correct verb here.

benfrankel (changes requested):
These are my last comments, then need to update the `.template` to match and I'll approve. Docs will be needed, but I won't block on that (so we can try out the action sooner).

.github/workflows/release.yaml
```diff
@@ -306,6 +310,13 @@ jobs:
           name: package-${{ matrix.platform }}
           retention-days: 1
 
+      - name: Upload binary to workflow artifacts for GitHub Pages (web)
+        if: ${{ env.is_platform_enabled == 'true' && matrix.platform == 'web' && inputs.deploy-to-pages }}
+        id: pages
```
benfrankel, line 315:
```suggestion
```
I believe this ID is the one that isn't needed.

Fjalar:
Just tested deployment on this branch, seems to be working

benfrankel (approved):
Looks good! I'll try it out after merging.

PR #472, opened 2025-11-14
Library copying for native platforms

Aceeri:
Adds library copying based on the platform that is being compiled.

Useful for copying things like steam sdks or other third party dlls.

For usage: have libraries in a folder that looks like this:
```
./         -- crate root
  lib/
    macos/
      libmylib.dylib
    linux/
      libmylib.so
    window/
      mylib.dll
      mylib.lib
```
and it'll be copied based on the platform to
```
./         -- application output directory
    app.exe
    assets/
    mylib.dll
    mylib.lib
```

Fixes https://github.com/TheBevyFlock/bevy_new_2d/issues/471

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,6 +54,9 @@ env:
   # The path to the assets directory.
   assets_path: assets
 
+  # The path to the libraries directory.
+  libs_path: lib
+
   # The itch.io project to deploy to in the format `user-name/project-name`.
   # There will be no deployment to itch.io if this is commented out.
   itch_page: the-bevy-flock/bevy-new-2d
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -172,6 +175,14 @@ jobs:
           if [ '${{ matrix.platform }}' != 'windows' ]; then
             RUSTFLAGS="${RUSTFLAGS:+$RUSTFLAGS }"'-Zshare-generics=y'
           fi
+          # Set up platform library rpaths
+          if [ '${{ matrix.platform }}' = 'linux' ]; then
+            RUSTFLAGS="${RUSTFLAGS:+$RUSTFLAGS }"'-C link-arg=-Wl,-rpath,$ORIGIN'
```
janhohenheim: nothing needed on Windows?

janhohenheim (approved):
Did not test this yet, but it *looks* correct from my limited understanding. Also, if this wasn't correct, it shouldn't break anything, so I'm in favor of merging :)
Will report at some point in the future if I can confirm that this works for Steam Audio haha

PR #476, opened 2026-01-18
port ui interaction to picking backend observers

olekspickle:
Let me know if it works for you or feel free to change

### DIFF
--- a/src/theme/interaction.rs
+++ b/src/theme/interaction.rs
@@ -3,10 +3,11 @@ use bevy::prelude::*;
 use crate::{asset_tracking::LoadResource, audio::sound_effect};
 
 pub(super) fn plugin(app: &mut App) {
-    app.load_resource::<InteractionAssets>();
     app.add_observer(apply_interaction_palette_on_click);
     app.add_observer(apply_interaction_palette_on_over);
     app.add_observer(apply_interaction_palette_on_out);
+
+    app.load_resource::<InteractionAssets>();
     app.add_observer(play_sound_effect_on_click);
     app.add_observer(play_sound_effect_on_over);
 }


src/theme/interaction.rs
```diff
@@ -85,3 +84,48 @@ fn play_on_click_sound_effect(
...
+    let Ok((palette, mut bg)) = palette_q.get_mut(hover.event_target()) else {
+        return;
+    };
+
+    *bg = palette.hovered.into();
+    commands.spawn(sound_effect(interaction_assets.hover.clone()));
+}
+
+fn on_out(
+    hover: On<Pointer<Out>>,
```
benfrankel:
```suggestion
    out: On<Pointer<Out>>,
```

I'm not sure which naming I prefer, but this is consistent.

src/theme/interaction.rs
```diff
@@ -85,3 +84,48 @@ fn play_on_click_sound_effect(
...
+) {
+    let Some(interaction_assets) = interaction_assets else {
+        return;
+    };
+    let Ok((palette, mut bg)) = palette_q.get_mut(hover.event_target()) else {
+        return;
+    };
+
+    *bg = palette.hovered.into();
+    commands.spawn(sound_effect(interaction_assets.hover.clone()));
```
benfrankel: I'm not sure I like handling both sound effects and background color in the same observer. As written, if a UI element has no `InteractionPalette` component, it will be unable to play interaction sound effects.

benfrankel (changes requested):
Will need to remove the now-unused old code too. Also I think the more explicit / specific observer and system naming style should be preferred, because it helps understand context when jumping around the codebase, reading error messages, and using schedule inspection dev tools.

src/theme/interaction.rs
olekspickle: hmm, that's true, I can separate sound observers, however that'll be a lot of the same function signatures...

Is there a way to do something like type alias to a bevy system arguments?

benfrankel: > Is there a way to do something like type alias to a bevy system arguments?

Not that I know of, but I think some boilerplate-type duplication is fine as long as the code is simple and readable. If there's something more clever that can also come in a later PR.

benfrankel (changes requested):

benfrankel (approved):
Only +4 lines while splitting a system into three observers is not bad. Mostly thanks to `If<T>`!

PR #478, opened 2026-01-31
Real CD

### LINKED ISSUE #138: Add workflow to keep release cache on `main` up-to-date
Workaround for: https://github.com/actions/cache/issues/1175.

Suggested here: https://stackoverflow.com/questions/69365200/github-actions-how-to-cache-dependencies-between-workflow-runs-of-different-bra.

janhohenheim:
Resolves #138

### DIFF
--- a/.github/workflows/release.yaml.template
+++ b/.github/workflows/release.yaml.template
@@ -135,6 +135,7 @@ jobs:
             echo "version=${{ inputs.version }}" >> "${GITHUB_OUTPUT}"
           else
             echo "version=$(cargo pkgid | sed 's/.*#//')" >> "${GITHUB_OUTPUT}"
+          fi
 
   # Build and package a release for each platform.
   build:


.github/workflows/release.yaml.template
```diff
@@ -62,10 +64,10 @@ env:
   {%- if itch_project != "" %}
   itch_page: {{itch_username}}/{{itch_project}}
   {%- else %}
-  itch_page: {{itch_username}}/{{project-name}}
+  itch_page: {{itch_username}}/{{project-name | kebab_case}}}
```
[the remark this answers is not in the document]
benfrankel: Sure, this is a more reasonable default if the user doesn't explicitly provide an `itch_project` value.

.github/workflows/release.yaml.template
```diff
@@ -99,7 +101,20 @@ env:
 
   # Enabling this only helps with consecutive releases to the same version (and takes up cache storage space).
   # See: <https://github.com/orgs/community/discussions/27059>.
+  # This is most useful in combination with `continuous_deployment_to_itch`
+  {%- if itch_username != "" %}
+  use_github_cache: true
```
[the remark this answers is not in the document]
benfrankel: Agreed, the cache hits are consistent which makes the builds very fast for fast iteration times. More specifically I'd say that caching is good for continuous releases, but bad for manual releases.

ChristopherBiscardi:
Needs a warning about potential private repo action cost

.github/workflows/release.yaml
```diff
@@ -92,7 +94,14 @@ env:
 
   # Enabling this only helps with consecutive releases to the same version (and takes up cache storage space).
   # See: <https://github.com/orgs/community/discussions/27059>.
-  use_github_cache: false
+  # This is most useful in combination with `continuous_deployment_to_itch`
+  use_github_cache: true
```
benfrankel: This enables caching for manual releases as well, which is a bad default. I'm not sure we should even be providing a knob that makes it easy to enable caching for manual releases.

My suggestion would be to either remove this env variable and always disable caching for manual releases and enable for CD, or keep this env variable but make it only affect CD.

benfrankel (changes requested):

.github/workflows/release.yaml
```diff
@@ -90,9 +92,15 @@ env:
...
-  use_github_cache: false
+  # Whether to deploy the newest version of the app on every push to `main` to itch.io.
+  # Requires `itch_page` to be set. Builds will ignore warnings.
+  # Warning: If you use private repositories, this will eat through your available worker minutes quickly.
+  continuous_deployment_to_itch: true
+
+  # Whether to deploy the newest version of the app on every push to `main` to GitHub Pages.
+  # Requires GitHub Pages to be set up. Builds will ignore warnings.
+  # Warning: If you use private repositories, this will eat through your available worker minutes quickly.
+  continuous_deployment_to_github_pages: false
```
benfrankel: For now this is fine, but if we add 1 more deployment target, I think we should condense this. Something like this:
```yaml
# Enable these to deploy the newest version of the app on every push to `main`.
# Make sure the enabled deployment targets are set up. Builds will ignore warnings.
# Warning: If you use private repositories, this will eat through your available worker minutes quickly.
continuous_deployment_to_itch: true
continuous_deployment_to_github_pages: true
```

.github/workflows/release.yaml
```diff
@@ -105,22 +113,31 @@ jobs:
...
       - name: Determine version number
-        id: tag
-        run: echo "ref=${GITHUB_REF#refs/*/}" >> "${GITHUB_OUTPUT}"
-    outputs:
-      # Use the input from workflow dispatch, or fall back to the git ref.
-      version: ${{ inputs.version || steps.ref.outputs.ref }}
+        id: version
+        # Use the input from workflow dispatch, or fall back to the Cargo.toml version
+        run: |
+          if [ -n "${{ inputs.version }}" ]; then
```
benfrankel:
```suggestion
          if [ -n '${{ inputs.version }}' ]; then
```
This is a template variable, so it may expand to something like this:

```sh
if [ -n 'v0.1.0-something*${quirky}' ]; then
```

We don't want any expansion to occur at this point, so single quotes should be used.

.github/workflows/release.yaml
```diff
@@ -105,22 +113,31 @@ jobs:
...
+            echo "version=${{ inputs.version }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "version=$(cargo pkgid | sed 's/.*#//')" >> "$GITHUB_OUTPUT"
```
benfrankel:
```suggestion
            echo "version=${{ inputs.version }}" >> "${GITHUB_OUTPUT}"
          else
            echo "version=$(cargo pkgid | sed 's/.*#//')" >> "${GITHUB_OUTPUT}"
```
Nit for consistency.

docs/workflows.md
```diff
@@ -32,6 +32,8 @@ The [CD workflow](../.github/workflows/release.yaml) will trigger on manual work
 - (Optional) Deploy to [itch.io](https://itch.io).
 - (Optional) Deploy to [GitHub Pages](https://docs.github.com/en/pages).
 
+If you have the `continuous_deployment_to_itch` env var enabled, it will also deploy the newest version of the app on every push to `main` to itch.io if the itch page is set up.
```
benfrankel:
```suggestion
If you enable `env.continuous_deployment_to_itch`, it will also deploy the newest version of the app to itch.io on every push to `main` (assuming the page is set up).
```
Nit

.github/workflows/release.yaml
```diff
@@ -215,7 +232,7 @@ jobs:
           targets: ${{ matrix.targets }}
 
       - name: Restore Rust cache
-        if: ${{ env.is_platform_enabled == 'true' && env.use_github_cache == 'true' }}
+        if: ${{ env.is_platform_enabled == 'true' && (needs.forward-env.outputs.continuous_deployment_to_itch == 'true' || needs.forward-env.outputs.continuous_deployment_to_github_pages == 'true') }}
```
benfrankel:
```suggestion
        if: ${{ env.is_platform_enabled == 'true' && github.event_name == 'push' }}
```
We should cache exactly when the platform is enabled and the build is for CD.

.github/workflows/release.yaml
janhohenheim: This defaults to `false` :) This is just the release workflow we use for this template itself.

benfrankel: Ah right, thanks :)

PR #480, opened 2026-02-05
Fix link in docs and add On<Release> interaction observer

olekspickle:
I found out I missed the release observer that resets color after button press

### DIFF
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -7,7 +7,7 @@ This template uses [GitHub workflows](https://docs.github.com/en/actions/using-w
 
 ## CI (testing)
 
-The [CI workflow](.github/workflows/ci.yaml) will trigger on every commit or PR to `main`, and it will:
+The [CI workflow](../.github/workflows/ci.yaml) will trigger on every commit or PR to `main`, and it will:
 
 - Check formatting.
 - Check documentation.
--- a/src/theme/interaction.rs
+++ b/src/theme/interaction.rs
@@ -43,7 +43,7 @@ fn apply_interaction_palette_on_release(
         return;
     };
 
-    *bg = palette.none.into();
+    *bg = palette.hovered.into();
 }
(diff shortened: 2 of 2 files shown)


benfrankel:
Wouldn't a pointer release change the bg color to hover, not none?

docs/workflows.md
```diff
@@ -7,7 +7,7 @@ This template uses [GitHub workflows](https://docs.github.com/en/actions/using-w
 
 ## CI (testing)
 
-The [CI workflow](.github/workflows/ci.yaml) will trigger on every commit or PR to `main`, and it will:
+The [CI workflow](../.github/workflows/ci.yaml) will trigger on every commit or PR to `main`, and it will:
```
benfrankel: Thanks for catching this.

janhohenheim (approved):
Good catch! Not sure about hover vs none, since I think you can hold press, move the cursor away, and then release? But yeah hover probably makes more sense

src/theme/interaction.rs
```diff
@@ -34,6 +35,17 @@ fn apply_interaction_palette_on_click(
...
 
+fn apply_interaction_palette_on_release(
+    click: On<Pointer<Release>>,
+    mut palette_query: Query<(&InteractionPalette, &mut BackgroundColor)>,
+) {
+    let Ok((palette, mut bg)) = palette_query.get_mut(click.event_target()) else {
+        return;
+    };
+
+    *bg = palette.none.into();
```
janhohenheim:
```suggestion
    *bg = palette.hovered.into();
```

janhohenheim (changes requested):

olekspickle:
I first made it hover, but then thought hover would take it over afterwards  anyway. Glad my first intuition turned out right :D

janhohenheim (approved):

PR #487, opened 2026-02-28
Fix interaction audio playing too often

Louis-Tarvin:
This is a bug I noticed while working on my Bevy Jam entry. The hover sound effect was playing whenever the mouse hovers over _any_ UI element, not just interactive ones. Similarly the click sound would play every time mouse1 is pressed.

Video demonstrating the bug:

https://github.com/user-attachments/assets/da0e8ae7-b94f-4250-81bf-87d267960356

The fix is to add a check that the entity the event triggered for has the `InteractionPalette` component

### DIFF
--- a/src/theme/interaction.rs
+++ b/src/theme/interaction.rs
@@ -88,17 +88,23 @@ impl FromWorld for InteractionAssets {
 }
 
 fn play_sound_effect_on_click(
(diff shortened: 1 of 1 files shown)


janhohenheim (approved):

janhohenheim:
Thx :)

PR #488, opened 2026-05-09
Fix lints url

Mclilzee:
Lints URL leads to 404 Not found page, doing a manual search led me to `bevy_cli/api/bevy_lint`


benfrankel (approved):

benfrankel (changes requested):
Can you update the URL in `Cargo.toml.template` as well?

benfrankel (approved):
Thanks :)

PR #489, opened 2026-06-27
Update to bevy 0.19

Fjalar:
Updates to bevy 0.19, adds the no longer implicit `"ui"` and `"audio"` feature flags, and fixes a few depreciation warnings. There seems to be a change in behaviour related to UI margin of a node which made the splash screen suddenly appear at the bottom despite `margin: UiRect::all(Val::Auto)` being set. I just removed that line, the default value for margin seems to put it in the center anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ license = "MIT OR Apache-2.0 OR CC0-1.0"
 
 [dependencies]
 # Use only features needed for a 2D game to improve compile time and size
-bevy = { version = "0.18", default-features = false, features = ["2d"] }
+bevy = { version = "0.19", default-features = false, features = ["2d", "ui", "audio"] }
 rand = "0.9"
 # Compile out low-severity logs to improve performance.
 # Remove these features if you want to profile your game with tracy.
--- a/src/screens/splash.rs
+++ b/src/screens/splash.rs
@@ -55,7 +55,6 @@ fn spawn_splash_screen(mut commands: Commands, asset_server: Res<AssetServer>) {
         children![(
             Name::new("Splash image"),
             Node {
-                margin: UiRect::all(Val::Auto),
                 width: percent(70),
                 ..default()
             },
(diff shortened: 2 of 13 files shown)


Fjalar:
I messed around with toolchain version and set bevy_lint to point to main in order to get the CI to pass, otherwise it complains about `cfg_select!` which has already been stabilized. The current commit passed CI on my fork, the error here is because some network thing. Is there a way to rerun CI in a pull request?

src/screens/splash.rs
```diff
@@ -55,20 +55,23 @@ fn spawn_splash_screen(mut commands: Commands, asset_server: Res<AssetServer>) {
...
-                    // `ImagePlugin::default_nearest()` is used for pixel art.
-                    settings.sampler = ImageSampler::linear();
-                },
-            )),
+            ImageNode::new(
+                asset_server
+                    .load_builder()
+                    .with_settings(
+                        // This should be an embedded asset for instant loading, but that is
+                        // currently [broken on Windows Wasm builds](https://github.com/bevyengine/bevy/issues/14246).
```
ollema, line 66: is this still broken?

ollema (commented):
one small question in the meantime, while waiting for a proper review from a maintainer

src/screens/splash.rs
Fjalar: I don't have a Windows PC to check. It's an older comment not from me, just shows up in the diff because of surrounding changes and indentation

Cargo.toml
atsjj, line 1: I gave your template a try and noticed that the `Cargo.toml.template` also needs to be updated as it currently generates with the `0.18` version of `bevy`. Probably also means the `Cargo.lock.template` needs updating too!

janhohenheim:
LGTM, thanks!

janhohenheim (approved):
Sorry for keeping y'all waiting for so long, and thanks for the PR <3

benfrankel:
I've pushed a `bevy-v0.18` tag pointing to the commit before this PR.

